PR #7, opened 2022-04-01
Add PathParser

Serabe:
This PR adds one of the options I came up to support dynamic segments in paths.

## How does this work (user land)?

The user defines a PathParser. It receives the same arguments as `DataSource.live_data/3` but it returns `{path, %{"segment_name" => value}}`. The second element of the tuple, is merged into params under `"path_params"`. Example:

```ex
defmodule MyApp.PathParser do
  def parse(_site, "blog", _params), do: {"blog", %{}}
  def parse(_site, "blog/authors", _params), do: {"blog/authors", %{}}
  def parse(_site, "blog/posts", _params), do: {"blog/posts", %{}}
  def parse(_site, "blog/authors/" <> author, _params), do: {"blog/authors", %{author: author}}
  def parse(_site, "blog/posts/" <> post, _params), do: {"blog/authors", %{post: post}}
end

defmodule MyApp.DataSource do
  def live_data(_site, "blog/authors", %{"path_params" => %{author: author}}=params) do
	%{author: Authors.get_by_slug(author)}
  end
  def live_data(_site, "blog/authors", _params) do
    %{authors: Authors.list_all()}
  end
end
```

## Other options

The other design (which I do like more but is not backwards compatible) is for `parse` to return `{String.t(), [String.t()]}`. In that case, the `DataSource` would look like:

```ex
defmodule MyApp.DataSource do
  def live_data(_site, {"blog/authors", [author]}, _params) do
	%{author: Authors.get_by_slug(author)}
  end

  def live_data(_site, {"blog/authors", _}, _params) do
    %{authors: Authors.list_all()}
  end
end
```

Which I think it looks better because it keeps the variable segments together with the page (which is actually a path).

### DIFF
--- a/lib/beacon/path_parser.ex
+++ b/lib/beacon/path_parser.ex
@@ -1,0 +1,11 @@
+defmodule Beacon.PathParser do
+  @behaviour Beacon.PathParser.Behaviour
+
+  def parse(site, path, params) do
+    get_path_parser().parse(site, path, params)
+  end
+
+  def get_path_parser do
+    Application.get_env(:beacon, :path_parser, Beacon.PathParser.Impl)
+  end
+end
--- a/lib/beacon/path_parser/behaviour.ex
+++ b/lib/beacon/path_parser/behaviour.ex
@@ -1,0 +1,3 @@
+defmodule Beacon.PathParser.Behaviour do
+  @callback parse(site :: String.t(), path :: String.t(), params :: map()) :: {String.t(), map()}
+end
--- a/lib/beacon/path_parser/impl.ex
+++ b/lib/beacon/path_parser/impl.ex
@@ -1,0 +1,5 @@
+defmodule Beacon.PathParser.Impl do
+  @behaviour Beacon.PathParser.Behaviour
+
+  def parse(_site, path, _params), do: {path, %{}}
+end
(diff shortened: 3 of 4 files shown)


bcardarella:
I like the direction as it is similar to what @TheFirstAvenger and I discussed briefly yesterday. I'd like to expand upon it. I think that Beacon itself should essentially be acting as its own routing system to a degree. In the Phoenix app's own router it greedily matches on the path that Beacon is mounted to and essentially hands off control to Beacon at that point (which is where I think you are heading too).

The dynamically compiled live actions are currently matching on the `slug` for the page. If instead we could define a DSL on the page author's side to allow for binary matching. For example, if we wanted to match on blog post IDs

```elixir
def live_action("/blog/" <> id, socket)
```

This is just pseduo code so it may not exactly align with the implementation. On the page creation UI the slug path would be: `/blog/:id`

We'd just need a way to parse this and at runtime produce the bytecode to do the proper binary matching.

Serabe:
Just a quick note on your comment, `/blog/:id` would not translate to `"/blog/" <> id`, that would be `/blog/*path`.

The problem I see with keeping these on the db is that all routing DSL depend on order, I'm not sure it's even possible not to depend on it. When having that in a file, it's easy to see what the actual order look like, but when this is in a db, it'd be hard to debug and in some cases almost impossible to debug. I.E. if we have `/blog/authors/:author/*resource` and `/blog/authors/:author/posts` it's easy to check that we are capturing the latter in the former, but having to take that into account when creating a page can be challenging at the very least.

Also, having a new dynamic path will most likely be for loading different data based on it (I think we can assume the `/hello/:name` examples will be the least), so modifying code does not look like an issue unless I'm missing something.

bcardarella:
@Serabe you're referring to the path matching DSL in the router. I'm referring to the actual pattern match on the function definition.

PR #12, opened 2022-07-09
Add tests setup

feliperenan:
While exploring the project, I noticed it still doesn't have a setup for
tests. This PR tries to add its initial setup.

### DIFF
--- a/test/beacon/pages_test.exs
+++ b/test/beacon/pages_test.exs
@@ -1,0 +1,69 @@
+defmodule Beacon.PagesTest do
+  use Beacon.DataCase, async: true
+
+  alias Beacon.Pages
+  alias Beacon.Layouts
+  alias Beacon.Components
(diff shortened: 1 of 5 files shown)


test/beacon/pages_test.exs
```diff
@@ -0,0 +1,69 @@
+defmodule Beacon.PagesTest do
+  use Beacon.DataCase, async: true
+
+  alias Beacon.Pages
+  alias Beacon.Layouts
+  alias Beacon.Components
+  alias Beacon.Stylesheets
+  alias Beacon.Pages.Page
+
+  defp create_page do
```
feliperenan, line 10: just copied it over from Readme to get the first example running. The focus of this PR is the setup, so not sure if we already want to do something more elaborated here.

bcardarella:
:+1: on this, I'll let @TheFirstAvenger weigh in before merge

PR #15, opened 2022-07-14
Add tests to Page Live

feliperenan:
This introduces our first test to PageLive which works as integration tests.


lib/beacon/loader.ex
```diff
@@ -3,6 +3,32 @@ defmodule Beacon.Loader do
 
   require Logger
 
+  defmodule Error do
```
feliperenan, line 6: @TheFirstAvenger I'm trying to improve the error message when the module is not yet available. Wdyt? I'm accepting suggestions to make this message even better 🙌🏻

TheFirstAvenger (approved):

Issue #22, opened 2022-07-29
LiveData not found error


TheFirstAvenger:
We need to have a nicer interface for when no pattern matches on the BeaconLiveData implementation. Right now it throws a function not found error.

PR #23, opened 2022-08-01
Improve live data error

feliperenan:
Closes https://github.com/BeaconCMS/beacon/issues/22


test/support/data_case.ex
```diff
@@ -22,4 +22,19 @@ defmodule Beacon.DataCase do
...
+  ### Examples
+
+      with_data_source(fn ->
+        # your test code in here...
+      end)
+  """
+  def with_data_source(datasource \\ DummyApp.BeaconDataSource, assert_func) when is_function(assert_func) do
+    Application.put_env(:beacon, :data_source, datasource)
+    assert_func.()
+    Application.delete_env(:beacon, :data_source)
```
TheFirstAvenger, line 38: Can we look at using Mox for the BeaconDataSource? I would like to avoid swapping env vars on the fly as it can get messy, and prevents us from using async tests.

feliperenan: True. I was trying to be simple here but forgot the messing with env prevents async tests. I will update it soon

feliperenan: actually, we don't need mox right now since I'm using the same datasource for both tests. We might need in the future once we get more tests that needs different data source, but we can live without it for the time being if that makes sense.

test/test_helper.exs
```diff
@@ -1 +1,3 @@
+Application.put_env(:beacon, :data_source, DummyApp.BeaconDataSource)
```
TheFirstAvenger, line 1: Lets move this to `config/test.exs` for now.

feliperenan: moved to `config/config.ex` where other configs are set atm. We don't need `config/test.exs` I guess 🤔

TheFirstAvenger: Let's add the standard `import_config "#{Mix.env()}.exs"` to `config.exs` and add the three `.exs` files, and put this config in `test.exs` so we don't cause runtime issues when we are in dev or prod mode, since that module is not compiled in those environments.

TheFirstAvenger:
Can we split out the two modules in DummyApp to their own files so it is easier to find where modules reside?

feliperenan:
@TheFirstAvenger I broke the dummy app into multiple files accordingly. I left it all in just one file because it is a "secondary app" but I agree it makes easier to find it that way.

TheFirstAvenger (approved):

PR #25, opened 2022-08-24
Add events to pages

### LINKED ISSUE #19: Handle Events
We need a way to handle LiveView events. The first pass at this will be to add an event schema to define what event mappings happen for a given page. It will have page, event pattern, and code fields. The code field will be Elixir code loaded into the handle_event function.

TheFirstAvenger:
Fixes #19.

This PR adds the concept of a `PageEvent`. Pages can have events, which for now define an event name and code. The code will have `event_params` and the `socket` in scope, and should return `{:noreply, socket}`.

See the seed in the README file walkthrough for an example.

Additional changes:
Safe Code is now disabled by default. Will re-enable once we have time to work on improving the SafeCode interaction.
Fix edge case with call_function_with_retry error handling.

### DIFF
--- a/lib/beacon/loader/page_module_loader.ex
+++ b/lib/beacon/loader/page_module_loader.ex
@@ -64,7 +64,7 @@ defmodule Beacon.Loader.PageModuleLoader do
       end
 
       """
-        def handle_event(#{path_to_args(path, "")}, #{event.event_name}, event_params, socket) do
+        def handle_event(#{path_to_args(path, "")}, "#{event.event_name}", event_params, socket) do
           #{event.code}
         end
       """
--- a/lib/beacon/loader/page_module_loader.ex
+++ b/lib/beacon/loader/page_module_loader.ex
@@ -3,6 +3,7 @@ defmodule Beacon.Loader.PageModuleLoader do
 
   alias Beacon.Loader.ModuleLoader
   alias Beacon.Pages.Page
+  alias Beacon.Pages.PageEvent
 
   def load_templates(site, pages) do
     page_module = Beacon.Loader.page_module_for_site(site)
@@ -10,7 +11,7 @@ defmodule Beacon.Loader.PageModuleLoader do
 
     # Group function heads together to avoid compiler warnings
     functions =
-      for fun <- [&render_page/1, &layout_id_for_path/1],
+      for fun <- [&render_page/1, &layout_id_for_path/1, &handle_event/1],
           page <- pages do
         fun.(page)
       end
@@ -34,7 +35,7 @@ defmodule Beacon.Loader.PageModuleLoader do
   end
 
   defp render_page(%Page{path: path, template: template}) do
-    if !Application.get_env(:beacon, :disable_safe_code, false) do
+    if !Application.get_env(:beacon, :disable_safe_code, true) do
       SafeCode.Validator.validate_heex!(template, extra_function_validators: Beacon.Loader.SafeCodeImpl)
     end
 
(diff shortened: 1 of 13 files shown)


lib/beacon/loader/page_module_loader.ex
```diff
@@ -56,6 +57,20 @@ defmodule Beacon.Loader.PageModuleLoader do
     """
   end
 
+  defp handle_event(%Page{path: path, events: events}) do
+    Enum.map(events, fn %PageEvent{} = event ->
+      if !Application.get_env(:beacon, :disable_safe_code, true) do
+        SafeCode.Validator.validate!(event.code, extra_function_validators: Beacon.Loader.SafeCodeImpl)
+      end
```
feliperenan, line 64: We may want to move this to some helper. So we could do something like:

```elixir
Beacon.SafeCode.validate!(event.code)
```

This would already check for the event so we don't need to duplicate this and it would be easier to enable/disable. Wdyt?

lib/beacon/pages.ex
```diff
@@ -251,4 +252,105 @@ defmodule Beacon.Pages do
...
+      {:error, %Ecto.Changeset{}}
+
+  """
+  def create_page_event(attrs \\ %{}) do
+    %PageEvent{}
+    |> PageEvent.changeset(attrs)
+    |> Repo.insert()
+  end
+
+  def create_page_event!(attrs \\ %{}) do
```
feliperenan, line 303:
```suggestion
  @docs """
  Same as create_page_event/1 but raises when there are validation errors.
  """
  def create_page_event!(attrs \\ %{}) do
```

priv/repo/migrations/20220823185841_create_beacon_page_events.exs
```diff
@@ -0,0 +1,17 @@
+defmodule Beacon.Repo.Migrations.CreateBeaconPageEvents do
+  use Ecto.Migration
+
+  def change do
+    create table(:beacon_page_events, primary_key: false) do
+      add :id, :binary_id, primary_key: true
+      add :code, :text
+      add :order, :integer
+      add :event_name, :text
```
feliperenan, line 9: should we force non-null for `event_name, order and `code`? 🤔

lib/beacon/loader/page_module_loader.ex
```diff
@@ -64,7 +64,7 @@ defmodule Beacon.Loader.PageModuleLoader do
       end
 
       """
-        def handle_event(#{path_to_args(path, "")}, #{event.event_name}, event_params, socket) do
+        def handle_event(#{path_to_args(path, "")}, "#{event.event_name}", event_params, socket) do
```
feliperenan, line 67: @TheFirstAvenger fixing the event pattern match in here. I guess it was supposed to match in the string `event_name`, right? Let me know if I'm missing something.

test/beacon_web/live/page_live_test.exs
```diff
@@ -39,21 +39,40 @@ defmodule BeaconWeb.Live.PageLiveTest do
...
+    Pages.create_page_event!(%{
+      page_id: page.id,
+      event_name: "hello",
+      code: """
+        {:noreply, Phoenix.LiveView.assign(socket, :message, "Hello \#{event_params["greeting"]["name"]}!")}
       """
     })
+
+    # Make sure events are loaded.
+    Beacon.Loader.DBLoader.load_from_db()
```
feliperenan, line 75: I wonder if we should also reload it after inserting the Event as we do for Pages 🤔

lib/beacon/loader/page_module_loader.ex
xn: Not only that, but all retrievals of `:disable_safe_code` are in `if !` form. The config value should be renamed to something like `validate_code_safety`, the boolean value flipped and the bang dropped.

xn: Also +1 on the moving this to a function, if I weren't clear. We should probably do this in a separate PR.

TheFirstAvenger: I'll create a follow up PR with this cleanup 👍

lib/beacon/loader/page_module_loader.ex
TheFirstAvenger: Yes, good catch!

lib/beacon/loader/page_module_loader.ex
TheFirstAvenger: @xn @felipe https://github.com/BeaconCMS/beacon/pull/27

PR #26, opened 2022-08-27
Improve tests

feliperenan:


lib/beacon/pages.ex
```diff
@@ -51,15 +51,21 @@ defmodule Beacon.Pages do
       {:error, %Ecto.Changeset{}}
 
   """
+  @reload_modules? Application.compile_env!(:beacon, :reload_modules?)
```
feliperenan, line 54: @TheFirstAvenger I'm not sure if switching envs is the right approach but I do think we should have a mechanism to reload it manually. That would be useful for tests at least.

To be fair, we don't really need it if you are not comfortable to take this decision right now since we don't have much tests - so let me know and we can close if that's the case :).

TheFirstAvenger: I would say we can defer the call on this one for now until/unless we run into a situation where the automatic reloading of modules becomes a real problem that we can't get around another way. Adding code to change functionality between test and other environments can lead to problems down the road, so lets cross this bridge when we get to it.

PR #29, opened 2022-09-16
Add cancel links to page manager forms

ryanwinchester:
Shouldn't need to use "back" buttons

### DIFF
--- a/lib/beacon_web/live/page_management/page_live/form_component.html.heex
+++ b/lib/beacon_web/live/page_management/page_live/form_component.html.heex
@@ -27,7 +27,7 @@
 
     <div>
       <%= submit "Save", phx_disable_with: "Saving..." %>
-      <%= live_patch "Cancel", to: "../pages" %>
+      <%= live_redirect "Cancel", to: "../pages" %>
     </div>
   </.form>
 </div>
--- a/lib/beacon_web/live/page_management/page_live/form_component.html.heex
+++ b/lib/beacon_web/live/page_management/page_live/form_component.html.heex
@@ -27,6 +27,7 @@
 
     <div>
       <%= submit "Save", phx_disable_with: "Saving..." %>
+      <%= live_patch "Cancel", to: "../pages" %>
     </div>
   </.form>
 </div>
--- a/lib/beacon_web/live/page_management/page_live/index.html.heex
+++ b/lib/beacon_web/live/page_management/page_live/index.html.heex
@@ -31,7 +31,7 @@ Last reload time: <%= @last_reload_time %>ms
         <td><%= page.template %></td>
 
         <td>
-          <span><%= live_patch "Edit", to: "page_editor/#{page.id}" %></span>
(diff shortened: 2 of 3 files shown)


lib/beacon_web/live/page_management/page_live/page_editor_live.html.heex
```diff
@@ -31,6 +31,7 @@
           <%= checkbox :opts, :publish, value: false %> Publish on Save
           <div>
             <%= submit "Save", phx_disable_with: "Saving..." %>
+            <%= live_patch "Cancel", to: "/page_management/pages" %>
```
TheFirstAvenger, line 34: Let's keep this relative so we can better support the page management residing in different paths
```suggestion
            <%= live_patch "Cancel", to: "../pages" %>
```

lib/beacon_web/live/page_management/page_live/form_component.html.heex
```diff
@@ -27,6 +27,7 @@
 
     <div>
       <%= submit "Save", phx_disable_with: "Saving..." %>
+      <%= live_patch "Cancel", to: "/page_management/pages" %>
```
TheFirstAvenger, line 30: Same
```suggestion
      <%= live_patch "Cancel", to: "../pages" %>
```

TheFirstAvenger (changes requested):

lib/beacon_web/live/page_management/page_live/page_editor_live.html.heex
ryanwinchester: Looks like `live_patch` on the `Edit` link in page_manager is not updating the url path to page_editor, so the `../` goes back to `/pages`.  I think we need to use `live_redirect` to get to page editor or use `./`?

ryanwinchester: https://user-images.githubusercontent.com/2897340/190716561-ef723275-f0d7-4d53-b48d-040c40bffc20.mov

ryanwinchester: Wait, isn't `live_patch` supposed to push a new state?
If you notice in the above video, I can't even use the back button

lib/beacon_web/live/page_management/page_live/index.html.heex
```diff
@@ -31,7 +31,7 @@ Last reload time: <%= @last_reload_time %>ms
         <td><%= page.template %></td>
 
         <td>
-          <span><%= live_patch "Edit", to: "page_editor/#{page.id}" %></span>
+          <span><%= live_redirect "Edit", to: "page_editor/#{page.id}" %></span>
```
ryanwinchester, line 34: `live_patch` was behaving oddly, probably since we're not linking within the same live view. I think we're supposed to use `live_redirect` for this.

TheFirstAvenger (approved):

PR #30, opened 2022-09-24
Add runtime compile of tailwind css

ryanwinchester:
This adds `Beacon.RuntimeCSS.recompile!/1` 

It recompiles and minifies/purges Tailwind CSS based on layout and pages from the DB.

### TODO (and to discuss):

- [x] Double check reading of templates/views in release (i.e. make sure the directories are correct)
- [x] Unit Test

**_NOTE_**: This adds the CSS to assigns in the module loader but not to any rendered pages. My main goal was getting CSS that could compile/recompile at runtime.

Also, there are also probably better ways to handle custom tailwind configs, but this is a starting point?


bcardarella:
@ryanwinchester we're starting to review this PR. Can you swap `mimic` our for `mox` as this is our preferred mocking library

ryanwinchester:
> To avoid changing application env vars on the fly, and also retain async on the tests, we should be able to set the `:css_compiler` mock in `test.exs` and then we can call the real `Beacon.CSSCompiler` directly in the test to test it. If we need to test it inline in another test, we can then `@runtime_css` and `expect` it in that specific test, maintaining async and with no env vars changing.

Yeah, thanks. That's a lot better. I never use `Mox` because I don't like all the extra ceremonial code just to be able to write some tests, and was following their docs.

AZholtkevych:
relates to #40

AZholtkevych:
resolves #40

AZholtkevych:
fixes #40

AZholtkevych:
Closes #40

lib/beacon/css_compiler.ex
```diff
@@ -0,0 +1,65 @@
...
+    0 = Tailwind.run(:runtime, ~w(
+      --config=#{config_file}
+      --input=#{input_file}
+      --output=#{output_file}
+      --minify
+    ))
+
+    output = File.read!(output_file)
+
+    cleanup!(config_file, output_file)
```
leandrocp, line 34: Just food for thought but we could leverage custom exceptions or raise with a custom message to expose better feedback, instead of returning something like MatchError or File.Error. Wdyt?

config/dev.exs
```diff
@@ -1 +1,8 @@
+import Config
 
+config :beacon, Beacon.Repo,
```
leandrocp, line 3: Is this config really necessary? 🤔

leandrocp (commented):
Hey @ryanwinchester an initial review to help getting the PR done.

lib/beacon/loader/layout_module_loader.ex
```diff
@@ -28,6 +29,7 @@ defmodule Beacon.Loader.LayoutModuleLoader do
 
   defp render_layout(%Layout{} = layout) do
     Beacon.Util.safe_code_heex_check!(layout.body)
+    runtime_css = RuntimeCSS.compile!(layout)
```
TheFirstAvenger, line 32: We need to call `render_layout` for the layout of a page whenever a page is updated.

config/dev.exs
ryanwinchester: It is if you want to run Beacon by itself for debugging in dev (e.g. `iex -S mix`)

lib/beacon/css_compiler.ex
leandrocp: Just to be clear, the current code looks good and there's no need to change it now :)

But eventually we'll need to provide feedback to users depending on the situation, similar to https://github.com/BeaconCMS/beacon/pull/35#issuecomment-1264392147

So we can move forward to get this PR merged and later discuss approaches to handle compilation errors. Makes sense?

lib/beacon/css_compiler.ex
```diff
@@ -0,0 +1,65 @@
...
+
+  alias Beacon.Components
+  alias Beacon.Layouts.Layout
+  alias Beacon.Pages
+
+  @behaviour Beacon.RuntimeCSS
+
+  @impl Beacon.RuntimeCSS
+  def compile!(%Layout{} = layout, opts \\ []) do
+    tmp_dir = System.tmp_dir!()
```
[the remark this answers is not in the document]
ryanwinchester: Updated https://github.com/BeaconCMS/beacon/pull/30/files#diff-d12b7007afa8a07707b6bf598cbb07d910acdbe84669cf03e27fdf5b75e1413fR61-R74

lib/beacon/css_compiler.ex
ryanwinchester: Generic runtime error with a bit of a better message for now: https://github.com/BeaconCMS/beacon/pull/30/files#diff-d12b7007afa8a07707b6bf598cbb07d910acdbe84669cf03e27fdf5b75e1413fR20-R36

leandrocp: Thanks! That's gonna help with debugging and identifying where we have to deal with compilation errors.

leandrocp:
Hey @ryanwinchester seems like it's missing the loading of `runtime_css` in `root.html.heex`. The `beacon.css` file is compiled and then injected into `layout_assigns` but seems like it's never loaded in the generated template. I was assuming it would be injected similar to other components here: https://github.com/BeaconCMS/beacon/blob/d0a75a593257103d267d7a0e3b02b5fccd1956e2/lib/beacon_web/templates/layout/root.html.heex#L7-L9

ryanwinchester:
> Hey @ryanwinchester seems like it's missing the loading of `runtime_css` in `root.html.heex`.

Yeah I had mentioned in previous discussions that I was adding it to the assigns but didn't know where those were being used and how to get the CSS into the actual rendered layout, but never really got an answer. So, just assumed it would be resolved later.

leandrocp:
> > Hey @ryanwinchester seems like it's missing the loading of `runtime_css` in `root.html.heex`.
> 
> @leandrocp Yeah I had mentioned in previous discussions that I was adding it to the assigns but didn't know where those were being used and how to get the CSS into the actual rendered layout, but never really got an answer. So, just assumed it would be resolved later 😂.

Ohh gotcha 😂
Yeah IMO that would be in `root.html.heex` but it's fine to add it in a following PR

leandrocp (approved):
I'm not sure about https://github.com/BeaconCMS/beacon/pull/30#discussion_r1046301407 but it LGTM.

Great work!

lib/beacon/loader/layout_module_loader.ex
TheFirstAvenger: Ahh yes, I didn't see it in this pr and forgot we were doing a "load the whole world" on publish_page. I added #76 to make that more focused.

TheFirstAvenger (approved):

leandrocp:
Thanks!!

PR #33, opened 2022-09-29
Make paths configurable

mayel:


TheFirstAvenger:
Thanks for adding this. Can you modify it so `Application.get_env` doesn't appear in the template, and we call Application.get_env from the `.ex` file instead?

mayel:
Does putting it in Util like this work for you?

TheFirstAvenger (approved):

mayel:
@TheFirstAvenger done

leandrocp:
Hey @mayel thank you so much for your PR. Can you please rebase with main?

PR #34, opened 2022-09-29
auto migrate on start

mayel:


TheFirstAvenger:
Thank you for this PR, however we don't want to automatically run migrations on startup. It becomes an issue when coordinating devops in multi-node deployments. It is better to leave the migrations to be manually run by whatever process the devops team determines is best (e.g. out of band between taking nodes down and bringing them up).

PR #35, opened 2022-09-29
skip invalid component instead of preventing app from starting

mayel:


bcardarella:
so this just prevents a runtime compilation error?

bcardarella:
I do agree that an invalid component shouldn't prevent the app from starting, but this is a more complex topic. I think that we should be logging these errors somewhere rather than suppressing them. We should be erring on the side of providing pages that are resilient and make best attempts to render but we need a way to notify the admin of the page that there was a compile error and why so they can resolve.

TheFirstAvenger:
Agreed @bcardarella, I would like to start with trying to catch these bad templates when they are being saved initially, as that gets the feedback to the right person at the right time. Maybe we can start with a changeset validation step that tries to compile the template being saved?

leandrocp:
Hey @mayel thanks for your PR. I'm closing this one in favor of https://github.com/BeaconCMS/beacon/issues/74 okay? Feel free to ping me if you want to work on that issue and need something.

PR #36, opened 2022-11-01
Up Elixir requirement to v1.14

leandrocp:
New Phoenix apps will soon require Elixir 1.14, see https://github.com/phoenixframework/phoenix/blob/04f9cc8d78c2435ed4f970467263c3bbb73025f9/installer/lib/mix/tasks/phx.new.ex#L339-L344

So it seems reasonable to set the same requirement for beacon. Wdyt?


TheFirstAvenger:
Thanks!

PR #61, opened 2022-12-08
Upgrade to Phoenix 1.7 and LiveView 0.18

leandrocp:
[Phoenix.Component](https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html) is a great improvement to Phoenix LiveView and we can leverage it as a building block for building pages if we consider that everything is a function component: layouts, pages, components (buttons, forms), 3rd party components, and so on. The trade-off is requiring Phoenix 1.7+ but that seems reasonable.

Keeping `Phoenix.View` for now to make it work with no much changes, but we could eventually move towards components only and remove the now soft-deprecated views.

Wdyt?

Refs:
* https://gist.github.com/chrismccord/00a6ea2a96bc57df0cce526bd20af8a7
* https://github.com/fly-apps/live_beats/commit/7bbce31bb9ecc246a411a3c92fe23c9af4e70f30
* https://github.com/phoenixframework/phoenix_live_view/blob/master/CHANGELOG.md#0180-2022-09-20


TheFirstAvenger (approved):
Nice job!

AZholtkevych:
FYI @leandrocp

PR #63, opened 2022-12-12
Add page helpers

### LINKED ISSUE #20: Views
We need a way to dynamically manage View helpers.

Context:
In DeadViews, we have a Controller, a Template, and a View, and helpers for the Template go in the View. In LiveView, these helpers go in the LiveView module. For Beacon, we don't have a LiveView module that the end user can modify, we dynamically load the template based on the path, and #19 will define the event handling.

Instead of having the developer hard code views, we should provide view helpers that are scoped to the template being rendered.

leandrocp:
This is a draft to get feedback and close #20

Added a public `dynamic_helper/2` function to route the fn call to the proper page module and also to avoid a cyclic dependency problem as the helper function isn't available yet when compiling the template.

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -1,6 +1,5 @@
-# Used by "mix format"
 [
   inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"],
   line_length: 150,
-  import_deps: [:ecto]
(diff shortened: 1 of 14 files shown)


TheFirstAvenger (approved):
Nice Job!

PR #72, opened 2022-12-15
Add dialyzer and fix spec warn

leandrocp:
We may not enforce dialyzer but eventually will need to debug and fix function calls, either to upstream (Phoenix, etc) or from user apps, so I believe it's worth keeping it.


lib/beacon_web/plugs/put_root_layout.ex
```diff
@@ -9,6 +9,6 @@ defmodule BeaconWeb.Plug do
 
   @impl Plug
   def call(%Conn{} = conn, _) do
-    Phoenix.Controller.put_root_layout(conn, {BeaconWeb.LayoutView, "root.html"})
+    Phoenix.Controller.put_root_layout(conn, {BeaconWeb.LayoutView, :root})
```
leandrocp, line 12: This is causing dialyzer errors on the user-project repo. Could be fixed by https://github.com/phoenixframework/phoenix/pull/5167

leandrocp:
> We should probably also add it to CI.

Sounds good.

What are your thoughts on using `@spec`? I was wondering if we could add to some functions along with `@doc`. My reasoning is to help explain how beacons works, even the internal modules, to help with contributions and future docs and guides.

TheFirstAvenger:
> What are your thoughts on using `@spec`? I was wondering if we could add to some functions along with `@doc`. My reasoning is to help explain how beacons works, even the internal modules, to help with contributions and future docs and guides.

I think it would be good to have the added checking, especially when we are getting into the lower level stuff that can be harder to grok what is going on, specs will be helpful.

leandrocp:
> > What are your thoughts on using `@spec`? I was wondering if we could add to some functions along with `@doc`. My reasoning is to help explain how beacons works, even the internal modules, to help with contributions and future docs and guides.
> 
> I think it would be good to have the added checking, especially when we are getting into the lower level stuff that can be harder to grok what is going on, specs will be helpful.

Great!

Issue #31, opened 2022-09-24
Asset Manager


bcardarella:
We need an interface for managing all assets that are uploaded to a site managed by Beacon.

I'd prefer that the asset manager does not permit for directories. It should be a flat file system that allows for the following

- [ ] upload new asset
- [ ] delete existing asset (soft delete, hard delete after a period of time... maybe 96 hours?)
- [ ] renaming of asset filename (don't allow for file extension rename)
- [ ] search - a live search to filter by name
- [ ] filter - filter by asset type (All, Image, Font, PDF, etc...)
- [ ] sort - by upload date, name, type
- [ ] used? - we should be able to show a list of all assets that are currently not in use
- [ ] which pages using - we should be able to show which pages are using a given asset, maybe even a grouping feature
- [ ] attributes for assets #84 

This should be a LiveView UI and available/unique per-site in Beacon

Assets should be stored in the database as binary data types

leandrocp:
@bcardarella and @TheFirstAvenger where the asset manager routes fit in the project? The current management namespace is `/page_management` so that would put the asset manager in `/page_management/assets` but that naming feels a bit off. Maybe renaming to something more generic like `management`, `admin`? Wdyt?

leandrocp:
> Assets should be stored in the database as binary data types

@bcardarella DB gives the portability we need but it's not that efficient and I was thinking how to serve those files without imposing infrastructure work on the final users, so an idea is a module that will hold binary content and serve it on a route:

On layouts and pages one can request an asset using a Beacon component:

```elixir
<BeaconWeb.MediaLibrary.image name="logo.jpg"/>
# or audio, pdf, and so on...
```

A controller will fetch and serve that asset:

```elixir
# controller
binary = Beacon.MediaLibrary.fetch_asset(name)

conn
|> Plug.Conn.put_resp_header("content-type", "image/jpg; charset=utf-8") # depends on the file
|> Plug.Conn.send_resp(200, binary)
```

And the module responsible for returning the binary, either from in-memory or from the DB:

```elixir
defmodule Beacon.MediaLibrary do
  # fetch from in-memory if available or fallback db
  def fetch_asset(name) do
  end
  
  # runtime compile functions holding binary content
  for {name, binary} <- Beacon.MediaLibrary.used_assets() do
    def "#{name}", do: binary
  end
end
```

[...]

PR #73, opened 2022-12-16
Migrate to layouts and components over views

leandrocp:
That's the second part of https://github.com/BeaconCMS/beacon/pull/61 and that will make Beacon look like a new Phoenix 1.7 app so we can move forward with https://github.com/BeaconCMS/beacon/issues/31


TheFirstAvenger (approved):

leandrocp:
@TheFirstAvenger thanks 🎉

PR #80, opened 2022-12-21
Add credo

lucas-cegatti:
Adds credo to the project and adds fixes according to its suggestions.

Two checks were removed:
- Credo.Check.Readability.ModuleDoc
- Credo.Check.Design.TagTODO

I also added it as part of the GitHub actions workflow.


leandrocp:
Thanks @lucas-cegatti !!

PR #86, opened 2022-12-23
Asset Manager

leandrocp:
Introduce `BeaconWeb.Admin.MediaLibraryLive` to manage assets and `BeaconWeb.Components` to display them.

Close the following tasks from issue https://github.com/BeaconCMS/beacon/issues/31:

- [x] upload new asset
- [x] delete existing asset (soft delete, hard delete after a period of time... maybe 96 hours?)
- [x] search - a live search to filter by name

Also added:

- [x] Assign `@beacon_attrs` that's injected into pages and used in components.
- [x] Route helper `beacon_asset_path/2` to resolve asset path for sites, also used in components.
- [x] `BeaconWeb.Components.image/1` to display an image. Other function components will be added.


leandrocp:
@AZholtkevych this PR will implement the core of https://github.com/BeaconCMS/beacon/issues/31 and extra features will be implemented in a following PR.

AZholtkevych:
Thanks a lot, @leandrocp . Added to #31

dev.exs
```diff
@@ -132,6 +132,9 @@ Beacon.Pages.create_page!(%{
     <h1 class="text-violet-900">Dev</h1>
     <p class="text-sm">Page</p>
     <%= my_component("sample_component", val: 1) %>
+    <div>
+      <BeaconWeb.Components.image beacon_attrs={@beacon_attrs} name="dockyard_1.png" width="200px" />
```
leandrocp, line 136: In order to resolve the asset path we need the `router`, see https://github.com/BeaconCMS/beacon/blob/90ec7814e05856be97717c76a81cf653ae3aba49/lib/beacon_web/components/components.ex#L24 and https://github.com/BeaconCMS/beacon/blob/90ec7814e05856be97717c76a81cf653ae3aba49/lib/beacon/router.ex#L159-L162

Another option was to inject beacon's assigns into socket private assigns and pass `@socket` instead, which is already available and has more data that we may need in the future but that seems overkill. Passing `@beacon_attrs` we make it explicit what that struct holds https://github.com/BeaconCMS/beacon/blob/90ec7814e05856be97717c76a81cf653ae3aba49/lib/beacon/beacon_attrs.ex and we can provide better docs too.

lib/beacon/router.ex
```diff
@@ -50,9 +50,16 @@ defmodule Beacon.Router do
         {session_name, session_opts, route_opts} = o = Beacon.Router.__options__(opts)
 
         live_session session_name, session_opts do
+          get "/beacon_assets/:asset", MediaLibraryController, :show
```
leandrocp, line 53: The endpoint to serve assets, each site gets its own. The trade-off is if a site is private (behind a plug or something) which could break this endpoint but if that ever happens we can provide a customization.

lib/beacon_web/components/components.ex
```diff
@@ -0,0 +1,27 @@
...
+  ## Examples
+
+      <BeaconWeb.Components.image beacon_attrs={@beacon_attrs} name="logo.jpg" width="200px" />
+  """
+  attr :beacon_attrs, BeaconAttrs, required: true
+  attr :name, :string, required: true
+  attr :class, :string, default: nil
+  attr :rest, :global
+
+  def image(assigns) do
```
leandrocp, line 22: More components to come. This one is to have something to prove that it works.

lib/beacon_web/controllers/media_library_controller.ex
```diff
@@ -0,0 +1,16 @@
+defmodule BeaconWeb.MediaLibraryController do
+  use BeaconWeb, :controller
+
+  alias Beacon.MediaLibrary
+  alias Beacon.MediaLibrary.Asset
+
+  def show(conn, %{"asset" => asset}) do
```
leandrocp, line 7: The lookup key `asset` comes from the component `name` attr: `<BeaconWeb.Components.image beacon_attrs={@beacon_attrs} name="dockyard_1.png" />`. That means this name has to be unique, which sounds okay.

TheFirstAvenger (approved):

PR #92, opened 2023-01-05
Tailwind integration

### LINKED ISSUE #85: Load compiled runtime CSS
That's a follow up of https://github.com/BeaconCMS/beacon/pull/30

The compiled `runtime_css` that is injected into layout assigns: https://github.com/BeaconCMS/beacon/blob/4c872bb90638130c6ff0a825d7b028367fe0b6b9/lib/beacon/loader/layout_module_loader.ex#L46-L48

Should be loaded in `root.html.heex` similar to the other resources: https://github.com/BeaconCMS/beacon/blob/4c872bb90638130c6ff0a825d7b028367fe0b6b9/lib/beacon_web/components/layouts/root.html.heex#L8-L11

### LINKED ISSUE #91: Tailwind integration for the BeaconCMS
As a PO, I would like to integrate Tailwind framework as a default CSS framework for the BeaconCMS, to use it for styling inside the Page Builder in the future

If it is only for Tailwind integration, then it is 5 hours

leandrocp:
Close #85 
Close #91 

- Move assets to /assets as expected by tailwind and esbuild
- Hide tailwind boilerplate config
- Inject js from host app
- Build dist js and css
- Load compiled css with tailwind
- Update dev.exs to leverage the same root layout
- Rename root.html.heex to runtime.html.eex to make it explicit it's used only for the user-facing sites

Note that we're making use of two separate templates now: `runtime` which is to be used by sites and will load dynamic resources (meta, stylesheets, etc) and `admin` which is to be used by `BeaconWeb.Admin`. With that we can optimize resources loading as the runtime template will be request way more often.


README.md
```diff
@@ -57,18 +57,7 @@ Beacon supports both.
       pool_size: 10
     ```
 
-7.  Configure Tailwind:
```
leandrocp, line 60: We can just inject an empty profile here https://github.com/BeaconCMS/beacon/blob/abb3aaade3adae66212a67b5ef1f2139349c5625/lib/beacon/css_compiler.ex#L20-L21

Eventually we'll support custom tailwind config but that should be good for now.

config/config.exs
```diff
@@ -6,9 +6,35 @@ config :beacon, :generators, binary_id: true
...
 
+config :esbuild,
+  version: "0.16.13",
+  default: [
+    args: ~w(js/app.js --bundle --target=es2020 --outdir=../dist/js),
+    cd: Path.expand("../assets", __DIR__),
+    env: %{"NODE_PATH" => Path.expand("../deps", __DIR__)}
+  ]
+
+config :tailwind,
```
leandrocp, line 17: This config is to be used by `dev.exs` and the other one is for dist.

lib/beacon_web/components/layouts/runtime.html.heex
```diff
@@ -9,10 +9,11 @@
     <title><%= page_title(assigns) %></title>
     <%= linked_stylesheets(assigns) %>
     <%= stylesheet_tag(assigns) %>
-    <%= if !dynamic_layout?(assigns) do %>
-      <link phx-track-static rel="stylesheet" href="/assets/app.css" />
-    <% end %>
```
leandrocp, line 14: @TheFirstAvenger what's the purpose of this conditional? If we're loading a user provided layout then it shouldn't load the default css provided by beacon? I'm wondering if that's still relevant because we're providing tailwind by default now.

lib/beacon_web/plugs/put_root_layout.ex
```diff
@@ -8,7 +8,18 @@ defmodule BeaconWeb.Plug do
...
+  def call(%Conn{} = conn, _opts) do
+    if admin?(conn) do
+      Phoenix.Controller.put_root_layout(conn, {BeaconWeb.Layouts, :admin})
+    else
+      Phoenix.Controller.put_root_layout(conn, {BeaconWeb.Layouts, :runtime})
+    end
   end
+
+  # TODO: provide a router macro, eg: beacon_site and beacon_admin
+  defp admin?(%{private: %{phoenix_live_view: {lv, _, _}}}) do
```
leandrocp, line 20: Relying on private API is definitely not ideal, I'd like to explore router macros to remove this block of code hence the `TODO`.

lib/beacon/css_compiler.ex
```diff
@@ -9,29 +9,37 @@ defmodule Beacon.CSSCompiler do
...
+  @template_tailwind_config_path Path.join("assets", "tailwind.config.js.eex")
+  @input_css_path Path.join(["assets", "css", "app.css"])
+
+  @external_resource @template_tailwind_config_path
+  @external_resource @input_css_path
+
   @impl Beacon.RuntimeCSS
   def compile!(%Layout{} = layout, opts \\ []) do
-    input_file = Path.join(:code.priv_dir(:beacon), "assets/beacon.css")
+    Application.put_env(:tailwind, :version, Beacon.tailwind_version())
```
TheFirstAvenger, line 20: I'm not a fan of libraries modifying app env vars on the fly. As a developer I would be upset if my app env vars changed without me changing them. Setting it if it isn't set already is still less than idea, but better than stomping on them. How about this?

```suggestion
    required_tailwind = Beacon.tailwind_version()
    case Application.get_env(:tailwind, :version, nil) do
      nil -> Application.put_env(:tailwind, :version, required_tailwind)
      ^required_version -> nil
      other -> raise "Beacon requires Tailwind version #{required_tailwind} but found #{other}"
    end
```

TheFirstAvenger (approved):

lib/beacon/css_compiler.ex
leandrocp: Yeah definitely it shouldn't change app's config silently :)

I've changed to just check if it's present instead of forcing a specific version.

PR #100, opened 2023-01-10
Loosen dependencies requirements

leandrocp:
Too strict will block apps to install beacon.


AZholtkevych:
@leandrocp thanks a lot, I see

PR #103, opened 2023-01-12
Implement Router Macros

### LINKED ISSUE #101: [Proposal] - Router Macros
### Router Macros

Introduce router macros to encapsulate route definition and allow customization on the site level.

#### `beacon_site(path, opts \\ [])`

Where `opts`:

- `live_socket_path` (optional) - same as LiveDashboard and Oban
- `data_source` (optional) - the current `config :beacon, :data_source` is global, one module for all sites. By having the config on the site level we can make it explicit which module belogs to each site or none if a site doesn't have a data source linked to it.
- `tailwind_config` (optional) - path to a `tailwind.config.js` file used to compile CSS for that site. Config on the site level so each site may have it's own customization, or just don't provide one to use the default config embedded in beacon. This config would be foundation to fix the issue https://github.com/BeaconCMS/beacon/issues/96

**Example:**

```elixir
beacon_site "/marketing"
```

*generates:*

```elixir
scope "/", BeaconWeb do
  live_session :beacon_marketing, session: %{"beacon_site" => "marketing"} do
    live "/marketing/*path", PageLive, :path
  end
end
```

Note that:
- The macro would generate the whole `scope` because the `BeaconWeb` namespace is essentially private, the user should not be able to change it, as `PageLive` is private.
- We can provide other `opts` to allow further customizations but that would be the minimal.
- `BeaconWeb.Plug` is not needed anymore, the root layout option would be injected in the `live_session` opts.

#### `beacon_admin(path, opts \\ [])`

The `path` here would be used to resolve the routes in the admin pages.

#### `beacon_api(path, opts \\ [])`

Similar to the other ones.

### References

- https://hexdocs.pm/phoenix_live_dashboard/Phoenix.LiveDashboard.Router.html#live_dashboard/2
- https://getoban.pro/docs/web/2.9.6/mounting.html

leandrocp:
Closes #101 

Introduce the following router macros:
- beacon_site
- beacon_admin
- beacon_api

This PR is also introducing some breaking changes:
- Deleted `BeaconWeb.Plug`, `BeaconWeb.PageManagement`, and `BeaconWeb.PageManagementApi` (no longer necessary)
-  Moved `config :beacon, :data_source` to router

I believe it's about time to start a CHANGELOG to document such changes, even tough a version wasn't released yet, wdyt?


TheFirstAvenger (approved):

AZholtkevych:
Thanks a lot, @leandrocp . Appreciate that

PR #112, opened 2023-01-17
Prefix layout meta tag fns with "layout_"

dukeweezo:
Following on from adding page meta tags, I renamed the layout meta tags functions from e.g. `meta_tags` to `layout_meta_tags`. Think I got everything, let me know if I missed something.


leandrocp:
Thanks @dukeweezo!

Issue #106, opened 2023-01-14
Custom live view socket path


leandrocp:
Currently it's connecting to the live socket provided by the host application at `/live` but that may not be case for every app and we want to support custom paths. For that an option `live_socket_path` is provided on `beacon_site` router macro but it has no effect currently (it's not implemented) https://github.com/BeaconCMS/beacon/blob/924202e757243b941154e576b18c13d447829ca0/lib/beacon/router.ex#L61

That value has to be fetched and injected into the runtime layout https://github.com/BeaconCMS/beacon/blob/924202e757243b941154e576b18c13d447829ca0/lib/beacon_web/components/layouts/runtime.html.heex so it can be passed to live socket https://github.com/BeaconCMS/beacon/blob/924202e757243b941154e576b18c13d447829ca0/priv/assets/js/app.js#L4-L7

Similar to how it's done on LiveDashboard.

lucas-cegatti:
@leandrocp I'll work on this issue.

PR #116, opened 2023-01-22
Feature/custom liveview socket path

lucas-cegatti:
Resolves https://github.com/BeaconCMS/beacon/issues/106

It replaces the default "/live" path using the option `live_socket_path` from `beacon_site` macro.

### DIFF
--- a/lib/beacon_web/components/layouts.ex
+++ b/lib/beacon_web/components/layouts.ex
@@ -26,6 +26,10 @@ defmodule BeaconWeb.Layouts do
   #{File.read!(phoenix_live_view_path) |> String.replace("//# sourceMappingURL=", "// ")}
   #{File.read!(beacon_js_path)}
   """
+  def render("app.js", %{conn: %{private: %{beacon: %{live_socket_path: live_socket_path}}}}) do
+    String.replace(@app_js, "/live", live_socket_path)
+  end
+
   def render("app.js", _assigns) do
     @app_js
   end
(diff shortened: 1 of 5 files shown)


lib/beacon_web/components/layouts.ex
```diff
@@ -26,6 +26,10 @@ defmodule BeaconWeb.Layouts do
   #{File.read!(phoenix_live_view_path) |> String.replace("//# sourceMappingURL=", "// ")}
   #{File.read!(beacon_js_path)}
   """
+  def render("app.js", %{conn: %{private: %{beacon: %{live_socket_path: live_socket_path}}}}) do
+    String.replace(@app_js, "/live", live_socket_path)
+  end
```
leandrocp, line 31: We can make it simpler by just returning the private value and injecting it into the template :)


```suggestion
  def live_socket_path(conn) do
    conn.private.beacon.live_socket_path
  end
```

And then add the following at https://github.com/BeaconCMS/beacon/blob/ec5e80c58891c62156298419f6b4b32ed3320eac/lib/beacon_web/components/layouts/runtime.html.heex#L2

```heex
<html lang="en" phx-socket={live_socket_path(@conn)}>
```

That works because:

1 - The value injected into `conn.private.beacon.live_socket_path` has always a default value or something custom defined by user, see https://github.com/BeaconCMS/beacon/blob/ec5e80c58891c62156298419f6b4b32ed3320eac/lib/beacon/router.ex#L61

2 - The LiveView JS client already expectes that html attr to find where to connect to, see https://github.com/BeaconCMS/beacon/blob/ec5e80c58891c62156298419f6b4b32ed3320eac/priv/assets/js/app.js#L4

leandrocp (changes requested):

lib/beacon_web/components/layouts.ex
lucas-cegatti: @leandrocp excellent, I've commited the recommended changes.

Thanks!

leandrocp:
Thanks @lucas-cegatti 🎉

Issue #119, opened 2023-01-25
Meta tag support


bcardarella:
We have layout meta tags but we also need per-page meta tags.

The app should essentially do a `Map.merge` on three sets of Meta tags:

1. site-wide - these are default tags that we always want to render unless there are higher priority overlapping tags
2. layout
3. page

On each page render we first get the Map of site tags, then merge the layout tags, then the page tags. Tags are considered unique based upon the `element` name and `id_key_name` and `id_key_value`. For example:

`<meta name="foorbar" content"abc">` and `<meta name="foobar" content="def">`. In this case they conflict and the later will overwrite the former.

`<meta name="foobar" content="abc">` and `<meta name="barbaz" content="def">` these do not conflict and will both render

`<meta name="foobar" content="abc">` and `<link name="foobar" content="def">` these do not conflict and will both render

`<meta name="foobar" content="abc">` and `<meta property="foobar" content="def">` these do not conflict and will both render

Some meta tags deviate from `name` being the `id_key`. For example, both are valid meta tags:

`<meta name="foobar" content="...">` and `<meta property="foobar"  content="...">` the `key` name needs to be accounted for. The data model could probably be:

`%{element: "meta", id_key: "name", id_key_value: "foobar", content_key: "content", content_value: "..."}`

bcardarella:
FWIW I think `<title>` should probably be included in this

leandrocp:
Hey @bcardarella per-page meta tags were implemented on the PR https://github.com/BeaconCMS/beacon/pull/109

From your description, we're currently missing:
1. site-wide meta tags
2. merge

We would move these default meta tags to `site.meta_tags` https://github.com/BeaconCMS/beacon/blob/c7f735f9b764dc0b136aa9a32d4901a80df91942/lib/beacon_web/components/layouts/runtime.html.heex#L4-L6 So every new site would bring a set of default meta tags that can be replaced on the layout or page level. Note that `<meta name="csrf-token" content={get_csrf_token()} />` is a special case that can't be replaced nor removed.

And then implement a `meta_tags(assigns)` function where we merge site-wide meta tags with layouts and pages as you described, ie: these lines would call `meta_tags(assigns)` to fetch and merge the 3 levels of meta tags: https://github.com/BeaconCMS/beacon/blob/c7f735f9b764dc0b136aa9a32d4901a80df91942/lib/beacon_web/components/layouts/runtime.html.heex#L8-L9

[...]

Issue #74, opened 2022-12-16
Core -> Route and Resource Loading: Handle runtime compilation errors


leandrocp:
See https://github.com/BeaconCMS/beacon/pull/35#issuecomment-1264392147 and https://github.com/BeaconCMS/beacon/pull/30#discussion_r1046000258 for more context.

Validate templates as they're saved or updated as suggested by @TheFirstAvenger: "start with trying to catch these bad templates when they are being saved initially, as that gets the feedback to the right person at the right time. Maybe we can start with a changeset validation"

leandrocp:
Hey @AZholtkevych I've assigned to myself just to reserve the issue because @alexandrexaviersm is working on it so we can make it to v0.1 ok?

AZholtkevych:
@leandrocp that's all good. Thanks a lot

PR #130, opened 2023-02-05
Merge site + layout + page meta tags

dukeweezo:
Here's a WIP of a PR to address issue #119 .

* Combined the `page_meta_tags` and `layout_meta_tags` fns, plus default sitewide  meta tags that are defined in `Beacon`. (If we'd like to provide the individual fns in future it should be easy).
* I chose a structure of meta tags of `[%{"name" => "home-meta-tag-one", "content" => "value"}, %{"name" => "home-meta-tag-two", "content" => "value"}]` which then gets converted to list of lists of tuples for ordering / ease of processing.

TBD
* Create overriding, presumably by order of succession (Page overrides Layout, which overrides Site), with protected `csrf-token` tag. Had this made but decided to redo it. Currently just doing an `Enum.concat`. 
  - As mentioned in the issue by @bcardarella, one option is to just leverage the overriding described here https://ogp.me/#array , allow duplicate name / property values, and do basic ordering of tags.
* Move and / or reuse some fns.

Just an early draft so feel free to point out any missteps!


lib/beacon/beacon.ex
```diff
@@ -21,4 +21,33 @@ defmodule Beacon do
...
+
+  def default_site_meta_tags() do
+    [
+      %{"charset" => "utf-8"},
+      %{"http-equiv" => "X-UA-Compatible", "content" => "IE=edge"},
+      %{"name" => "viewport", "content" => "width=device-width, initial-scale=1"},
+      %{"name" => "csrf-token", "content" => Phoenix.Controller.get_csrf_token()}
+    ] 
+    |> Enum.map(fn tags -> Map.to_list(tags) end) 
+    |> order
```
bcardarella, line 33: can you add the `()`

schrockwell:
Should we also roll the page title overrides into this API, as well? From the outside, it makes sense to group them together because they are also accessed and rendered in the `<head>` tag.

bcardarella:
Pretty much everything in the `<head>` tag I think should probably be encapsulated in a single place. But how we represent the different tags may conflate the API. As the `<title>` tag doesn't set its content to a property but the actual tag content. That would be a very different rendering than the `<meta>` tags. But we would want the same hierarchy of page titles. So:

`site title < layout title < page title`

This would mean that we'd always ensure that pages have a title regardless of it being set. This is a good practice IMO.

*But!* I am going back and forth on the meta tag overriding. The architecture that I laid out in #119 was informed by how I implemented meta tags in Ember for the current production dockyard.com. It is worth acknowledging that I didn't reference the actual specification for RDFa or Open Graph. Considering Beacon is intended to be a general purpose CMS and not just for dockyard.com I think we have a responsibility to implementing to the specification and not just our needs.

The good news is I don't think this PR would require too many changes. Or maybe none at all. What I'm getting at is let's layout the specification and discuss how best to align with the spec.

Based upon the spec meta tags can have duplicate `name/property` fields per page. It is up to the consumer to decide if they're all used as a list or if the first one present in the html is used. However, I looked around several very large content sites and I cannot find an example of this being used in production. I'm OK keeping the current implementation but I wanted to provide an opportunity to explore the alternative first before we mint this.

bcardarella:
Also, something that LiveView doesn't currently do is update `<head>` content during live navigation. There is a `<title>` escape hatch but many crawlers are now starting to execute JavaScript rather that rely on SSR. This means that meta tag content won't be updated during live navigation events and SEO would suffer.

leandrocp:
IMO we should aim to keep separated APIs for each html tag, following the spec as much as possible. Both `meta` and `title` are under the same [metadata content](https://html.spec.whatwg.org/#metadata-content-2) category but their structure and requirements differs, and the tags serve different purpose semantically. This way the api is simple and flexible for future changes, and both runtime.html.heex and admin.html.heex can consume these layout functions as needed.

bcardarella:
@leandrocp I meant from the editing perspective but yes the data model should be represented differently

leandrocp:
@bcardarella what about applying diffs on `head`? Should we try pushing a proposal to LV to support something similar to `live_title`? That's gonna be required for SEO purposes, not only for Beacon.

dukeweezo:
Hey, just did some [changes to the PR](https://github.com/BeaconCMS/beacon/pull/130/commits/5e41509eccc4970ffc039f00010bbf396645cf0a). 

- Corrected logic to reorder attributes, even tags with 3+ attributes, (with tests) 
- Made tags show up in correct order of Site < Layout < Page
- Also made it filter out a `csrf-token` tag from Page and Layout if this gets included
Screen Shot 2023-02-07 at 5 21 27 PM
Screen Shot 2023-02-07 at 5 13 24 PM




Interesting discussion! On the meta tags, my instinct would be to allow duplicate tags and order them like above, according to the Open Graph spec, while perhaps providing a disclaimer / warning of duplicate tags in the UI to the user. 

(Will write a test to confirm the correct order of the tags in the next day or so).

bcardarella:
@leandrocp I spoke with Jose today about this. He had a proposal that I'm not entirely sold on. The idea would be to render the meta tags within a `<template>` tag of a LV page. The tag would be hidden via CSS then adding an EventListener on `phx-hook` to replace the tags in the `<head>` with the content of that `<template>` tag. This doesn't feel great as it would mean we'd have to control for JS pages renders vs SSR. SSR should always render the meta tags for the page. I'd like to find a single solution

bcardarella:
I think that @TheFirstAvenger and I spoke about this very early on in Beacon dev and I forget his counter-point so I'll bring it up again. Why can we not have LV control the entire document? I know the root template is not LV enabled by why not make the root layout? Just have `data-phx-*` attributes on `<html>` ? That way you could have the entire document under the control of LV.

leandrocp:
@bcardarella regarding "an EventListener on phx-hook to replace the tags in the <head> with the content of that <template> tag".

That would work as long as JS is enabled on users' browsers, that's one reason why we need SSR. Another reason is that we need to be extra careful to make sure the JS execution doesn't fail. If either one of those fails, we still render outdated meta tags.

In terms of SEO, updating meta tags on the client side seems to be okay given how some other popular libs work, eg:
* https://github.com/nuxt/vue-meta/blob/8c19e183495c830ae8a61e2d925f2948538e1864/src/client/load.js
* https://github.com/s-yadav/react-meta-tags/blob/5a77f92d1526734d3713ff7d4db44c3edb9f3328/src/meta_tags.js

> Just have data-phx-* attributes on <html> ?

Following the current LV patterns, I'd say that having a `live_meta_tag` or `live_meta_tags` would make more sense (depends on the api and how the state is handled).

leandrocp:
@dukeweezo

> On the meta tags, my instinct would be to allow duplicate tags and order them like above, according to the Open Graph spec, while perhaps providing a disclaimer / warning of duplicate tags in the UI / console to the user.

IMO we should keep it simple for now. Let users be responsible for defining the correct meta tags and there's no need to emit warnings, just a note on docs would be fine. We can optimize it later as we learn how meta tags are being used.

TheFirstAvenger:
> I think that @TheFirstAvenger and I spoke about this very early on in Beacon dev and I forget his counter-point so I'll bring it up again. Why can we not have LV control the entire document? I know the root template is not LV enabled by why not make the root layout? Just have `data-phx-*` attributes on `<html>` ? That way you could have the entire document under the control of LV.

@bcardarella Here is the internal [Slack Conversation](https://dockyard.slack.com/archives/C02NCP1799C/p1638891315010400) we had on it. The root issue is that live redirects won't update the `<head>` tag, so you will get the right meta tags on the first render, but clicking around won't update the meta tag. This isn't as big of a deal, but there are issues in regards to other page-specific tags in the head, such as setting different styles on different pages won't take or won't unload on live-redirect, so you get non-deterministic behavior for hard refresh vs live navigation.

lib/beacon_web/components/layouts.ex
```diff
@@ -70,24 +70,68 @@ defmodule BeaconWeb.Layouts do
...
+    site_meta_tags = site_get_meta_tags()
+
+    layout_meta_tags = layout_get_meta_tags(assigns)
+
+    page_meta_tags = page_get_meta_tags(assigns)
+
+    page_meta_tags
+    |> merge_tags(layout_meta_tags)
+    |> filter_out_attr_value("csrf-token")
+    |> merge_tags(site_meta_tags)
```
leandrocp, line 83: Concatenation does the job here:

```suggestion
    site_meta_tags = site_get_meta_tags()
    layout_meta_tags = layout_get_meta_tags(assigns)
    page_meta_tags = page_get_meta_tags(assigns)

    page_meta_tags = Enum.reject(page_meta_tags, & &1["name"] == "csrf-token")
    site_meta_tags ++ layout_meta_tags ++ page_meta_tags
```

The bottom line is to allow duplicate meta tags because that's indeed allowed, for eg `og` properties and also regular meta tags like the `http-equiv` directive. And leverage the list order.

leandrocp (changes requested):
Hey @dukeweezo I believe we can simplify the implementation.

Ordering isn't necessary because 1) we can rely on list ordering; and 2) we should allow duplicated meta tags anyway.

dukeweezo:
@leandrocp Thanks for the feedback!

> Ordering isn't necessary because 1) we can rely on list ordering; and 2) we should allow duplicated meta tags anyway.

Just to clarify, that code is for ordering attributes, not meta tags. So, for example, making sure the `property` or `name` (and `charset`, etc.) attribute always precede the `content` attribute, and the grouping of `property[ / name, etc. ...] content` always precedes other attributes (if they're ever needed). For an example [see](https://github.com/BeaconCMS/beacon/pull/130/commits/5e41509eccc4970ffc039f00010bbf396645cf0a#diff-a5b9ca0866de29f3449862c9f1b7009a8f50543c21efe6f890e99e253e8f400b). I wanted to convert from map / order the attributes in an `Ecto.Type` to guarantee tuple order as early as possible. Granted there are other ways to ensure the order.

This was partly to create a strong guarantee for merging logic described in the original issue (which has now admittedly been swapped for straightforward concatenation with duplication), but potentially also for future functions interacting with meta tag attributes. Also, not sure if the order `<meta property[ / name, etc. ...] content[... other attrs]` is required by any specification, but it seems like good practice.

Regarding the concatenation / filtering, I incorporated `merge_tags` and `filter_out_attr_value` functions in case we were contemplating different merge strategies and wanted to reuse functions, but sounds like the concat operator, etc. is all we need right now. :) (Good spot about `Enum.reject`, that's more succinct.)

leandrocp:
@dukeweezo 

> that code is for ordering attributes, not meta tags

No need to worry about that, any order will work :) 

> sounds like the concat operator, etc. is all we need right now

That's right 👍🏻 

--

In short, even if someone inputs a custom name or an invalid attribute like `%{"name" => "custom-name, "content" => "whatever", "another-attr" => "value"}` then it should be rendered, we can't enforce strict rules at this moment. Note that custom names are allowed on the spec - https://html.spec.whatwg.org/multipage/semantics.html#other-metadata-names

Also some meta tags have more than the pair name/content attrs, for eg: `<meta name="theme-color" content="#3c790a" media="(prefers-color-scheme: dark)">`

So given a tag:

```elixir
tag = %{"name" => "custom-name", "content" => "whatever", "another-attr" => "value"}
```

You can render it as:

```elixir
List.to_string(["<meta ", Enum.map_join(tag, " ", fn {key, value} -> [key, "=", value] end), " />"])
```

Which also works for tags like `%{"charset"=> "utf-8"}`

dukeweezo:
@leandrocp Yeah the `order_attributes/1` wasn't to enforce strict rules or anything. It was made to work for the cases you mentioned (including meta tags with invalid or extra attributes other than name/content). It was just to guarantee attribute ordering. 

For example, in the code you provided it produces
`"<meta another-attr=value content=whatever name=custom-name />"`. Changing the `another-attr` to `zebras` in the map, you get `"<meta content=whatever name=custom-name zebras=value />"` (so the produced order is more or less, but not necessarily guaranteed, alphabetical).

Definitely a minor issue so I won't take up anyone's time. But if nothing else having a guaranteed structure makes it easier for the tests like `render meta tags for a layout` (you can just do a simple `assert` against a string sigil instead of regex).

It's also potentially useful if you want a function that does pattern matching against the tag and you want a guarantee to do something like `[{name, val}, {content, val} | other_attrs]`.

leandrocp:
> No worries if it gets nixed

@dukeweezo I'd like to get it merged, keep going :)

Regarding tests, I agree it would make it easier but it's not worth the added complexity. A couple of regexs on tests are fine 👍🏻

dukeweezo:
OK, rewrote those tag merge / join fns similar to above, got rid of the attribute ordering, and added tests using regex. The regex in `fixtures.exs` is little verbose, but I opted for simple and verbose over some generating function that would add complexity.

Screen Shot 2023-02-11 at 10 39 42 AM

leandrocp:
Hey @dukeweezo seems like it's almost done but please let me know if you need anything.

dukeweezo:
@leandrocp thanks for asking! Yes, I'm not quite sure how to get rid of the error `** (CompileError) nofile:1: cannot define module BeaconWeb.LiveRenderer.Layout6A217F0F7032720EB50A1A2FBF258463 because it is currently being defined in nofile:1`--probably something obvious.

leandrocp (changes requested):
Hey @dukeweezo I'm leaving comments to help you get that error fixed.

Also you'll need to update the meta tags values in `dev.exs` and `templates/install/seeds.exs` - just grep and you'll find it :)

dukeweezo:
@leandrocp Like I said, something obvious... thanks! Thought I had made migrations but obviously not. (That's definitely a good idea not to just drop all the meta tags for everyone who isn't using a seeds file)

I still fail some tests with the error I mentioned above, though not consistently. Maybe that's not an issue though

leandrocp:
Hi @dukeweezo 

> not to just drop all the meta tags for everyone

It's fine to introduce breaking changes at this pre-release version but it's a good practice to avoid it if possible.

> I still fail some tests

Don't worry, those will be fixed in https://github.com/BeaconCMS/beacon/pull/134. It's basically a race condition.

> set up a constraint at the schema / database 

Let's not have it for now because we still need to work on validating runtime errors, see https://github.com/BeaconCMS/beacon/issues/74 - and that constraint would be another validation in the pipeline but we still need to think about how to better provide feedbacks in case of errors.

leandrocp:
@dukeweezo I believe this PR is good to go. Do you wanna push more commits before making it ready for review? Thanks!

dukeweezo:
> @dukeweezo I believe this PR is good to go. Do you wanna push more commits before making it ready for review? Thanks!

If it looks good to you guys, I think it's all ready. (That response above is much appreciated. Answers some lingering questions I had.)

leandrocp:
> > @dukeweezo I believe this PR is good to go. Do you wanna push more commits before making it ready for review? Thanks!
> 
> If it looks good to you guys, I think it's all ready. (That response above is much appreciated. Answers some lingering questions I had.)

To give you a bit more context, we're using Beacon to build the new dockyard.com website and your contribution is much appreciated and will be of good use. Which also means that I can make small fixes as I test on the bigger repo that has more usage cases for this feature 👍🏻

dukeweezo:
> > > @dukeweezo I believe this PR is good to go. Do you wanna push more commits before making it ready for review? Thanks!
> > 
> > 
> > If it looks good to you guys, I think it's all ready. (That response above is much appreciated. Answers some lingering questions I had.)
> 
> To give you a bit more context, we're using Beacon to build the new dockyard.com website and your contribution is much appreciated and will be of good use. Which also means that I can make small fixes as I test on the bigger repo that has more usage cases for this feature 👍🏻

That's good to know. Weren't aware you guys were doing a redo of dockyard.com. Happy to contribute on this when I can! I admit I have personal ulterior motives, because I've been wanting a full-fledged CMS in Elixir for a while now. Thought of making my own but this is much better. 😃

leandrocp:
Thanks @dukeweezo !

PR #132, opened 2023-02-05
Inject beacon seeds into ecto.setup alias

hubertpompecki:
Updates the installation mix task so that beacon's seeds are run on `mix setup`.


leandrocp:
Thanks @hubertpompecki 🎉

PR #134, opened 2023-02-09
Represent site name identifier as `atom` and introduce `Beacon.Config`

leandrocp:
A bit of context first. Config used to be global, same config for all sites as:

```elixir
config :beacon, data_source: MyApp.BeaconDataSource
```

Worked great to get it up and running but you can see the limitation. A `site` is essentially a tenant, where its data _and_ config should be isolated. Then I moved config to `beacon_site` on router as:

```elixir
beacon_site "/my_site", name: "my_site", data_source: MyApp.BeaconDataSource
```

Solved the tenancy problem but that won't work for all kinds of configuration, especially for values that are required very early on the pipeline like `tailwind_config` which is being introduced on https://github.com/BeaconCMS/beacon/pull/129.

~The next iteration, implemented in this PR, is reverting back to `config` but keeping the tenancy flexibility along with the `beacon_site` macro:~

```elixir
# config.exs or runtime.exs
config :beacon, otp_app: :my_app

config :my_app, Beacon,
  sites: [
    my_site: [
      data_source: BeaconDataSource,
      # ... all other configs
      # if all are optional, this `config` is not even required
    ]
  ]

# router
beacon_site "/my_site", name: :my_site
```

Actually the final version is passing the options on the Beacon supervisor, eg: https://github.com/BeaconCMS/beacon/blob/25b62c9d95e9a3ed099a3fdd22a59dc64706d620/lib/beacon/beacon.ex#L22-L41

In short, a project may have multiple Phoenix apps (in case of umbrella apps) and each app may have multiple sites, each with its own data and configuration, except for `Repo` but that's another case.

But that brings another issue, the site `:name` identifier data format. I've changed to `atom` instead of `binary`, it makes more sense for a identifier and we can explore using module-based supervisors as well. It introduces a breaking change but it's more important to get it right now than later.

The only trade-off I can see is asking users to pay attention to use the same `:name` on `beacon_site` and `config`, and that was another reason to change it to `atom`, as config common pattern is to use keywords.

As a side note this design was inspired on Oban. It's simple but flexible enough to introduce more configs as needed without breaking changes in the future.


leandrocp:
Hey @TheFirstAvenger and @schrockwell this is a draft but most of it is done and I appreciate an early review and feedback.

TheFirstAvenger:
I like it. The only suggestion I have is that instead of:
```elixir
beacon_site "/my_site", name: :my_site
```
I think it should be:
```elixir
beacon_site "/my_site", site: :my_site
```
since `:my_site` is now the actual site identifier atom and not just a string name.

leandrocp:
> I like it. The only suggestion I have is that instead of:
> 
> ```elixir
> beacon_site "/my_site", name: :my_site
> ```
> 
> I think it should be:
> 
> ```elixir
> beacon_site "/my_site", site: :my_site
> ```
> 
> since `:my_site` is now the actual site identifier atom and not just a string name.

Makes sense.

leandrocp:
Hey @TheFirstAvenger I had to iterate once more (https://github.com/BeaconCMS/beacon/pull/134/commits/8c9ad9422b5cad4d8bdafcca15e8559003bd62a8) to get it working on Umbrella apps but turns out it made me realize that we can control how Beacon itself and each site is loaded. The bottom line is passing all options to the main Beacon supervisor and then starting one supervisor for each site where it holds the configuration on Registry and also starts one Loader.Server for each site under the site supervisor. This model is similar to Oban, Broadway and others. Some benefits:

1. We can control when the Loader.Server gets started. On an Umbrella app that means loading only on `web`, a side effect is not reloading all modules when running seeds, see:

Running `time mix ecto.reset` on dockyard website:

Before: `mix ecto.reset  0.52s user 0.77s system 0% cpu 3:01.40 total`
After: `mix ecto.reset  0.37s user 0.35s system 11% cpu 6.306 total`

Correct If I'm wrong but I believe we don't need module reloading to run seeds.

2. Isolate module loading per site, so a failure on one site won't impact others.

3. All options are passed in the same place as it's stored in the same place too.

-----------

An example of how it looks like for the final user https://github.com/BeaconCMS/beacon/blob/8c9ad9422b5cad4d8bdafcca15e8559003bd62a8/lib/beacon/beacon.ex#L22-L41

-----------

The main changes/files on that commit:

- Pass options, eg: https://github.com/BeaconCMS/beacon/blob/8c9ad9422b5cad4d8bdafcca15e8559003bd62a8/test/test_helper.exs#L17-L30

- What gets started on dep loading https://github.com/BeaconCMS/beacon/blob/8c9ad9422b5cad4d8bdafcca15e8559003bd62a8/lib/beacon/application.ex#L11-L15

_I wish we could isolate PubSub and Repo per site as well but I'm aware of the migrations issue, that's for another time._

- Start a supervisor for each site https://github.com/BeaconCMS/beacon/blob/8c9ad9422b5cad4d8bdafcca15e8559003bd62a8/lib/beacon/beacon.ex#L59

- Load from DB per site https://github.com/BeaconCMS/beacon/blob/8c9ad9422b5cad4d8bdafcca15e8559003bd62a8/lib/beacon/loader/db_loader.ex#L20

_Site is essentially a tenant._

leandrocp:
Hey @TheFirstAvenger this should be good for a final review, I'm just testing edge cases on existing sites. It's a bit long so let me know if you need more context.

TheFirstAvenger:
@leandrocp thoughts on using an [`Ecto.Enum`](https://hexdocs.pm/ecto/Ecto.Enum.html) for the site instead of a custom type?

leandrocp:
> @leandrocp thoughts on using an [`Ecto.Enum`](https://hexdocs.pm/ecto/Ecto.Enum.html) for the site instead of a custom type?

I'm not sure how that could work because `Ecto.Enum` requires pre-defined values, right?

TheFirstAvenger:
@leandrocp ahh yes, good point. The PR looks great, nice job!

TheFirstAvenger (approved):

PR #144, opened 2023-02-24
Fix query error: comparison with nil is forbidden - /admin/pages/new

alexandrexaviersm:
I get the following error after going to the `http://localhost:4000/admin/pages/new` route (after running the project for the first time)
```elixir
** (ArgumentError) comparison with nil is forbidden as it is unsafe. If you want to check if a value is nil, use is_nil/1 instead
    (ecto 3.9.4) lib/ecto/query/builder.ex:1045: Ecto.Query.Builder.not_nil!/1
    (beacon 0.1.0-dev) lib/beacon/layouts.ex:27: Beacon.Layouts.list_layouts_for_site/1
    (beacon 0.1.0-dev) lib/beacon_web/live/admin/page_live/form_component.ex:75: BeaconWeb.Admin.PageLive.FormComponent.assign_layouts/1
```

This problem is happening because we tried to load a list of **Layouts** based on a **site**, but at this moment we still don't know which **site** to use, so we call `Layouts.list_layouts_for_site(nil)` and the error above happens.

So before trying to load the `list_layouts_for_site` without a site, we need to load the existing sites and display them on the form. 

ezgif com-gif-maker

**Changes:**
- Change the form text input field to type "select" and display the existing sites in the form
- Create a function to load existing sites - `list_distinct_sites_from_layouts/0`
- Add `assign_layouts/2` to assign the available **Layouts** according to the selected **site**

### DIFF
--- a/lib/beacon/layouts.ex
+++ b/lib/beacon/layouts.ex
@@ -21,7 +21,7 @@ defmodule Beacon.Layouts do
     Repo.all(Layout)
   end
 
-  def list_layouts_for_site(site) when not is_nil(site) do
+  def list_layouts_for_site(site) do
     Repo.all(from l in Layout, where: l.site == ^site)
   end
 
--- a/lib/beacon/layouts.ex
+++ b/lib/beacon/layouts.ex
@@ -21,6 +21,8 @@ defmodule Beacon.Layouts do
     Repo.all(Layout)
   end
 
+  def list_layouts_for_site(nil), do: []
+
   def list_layouts_for_site(site) do
     Repo.all(from l in Layout, where: l.site == ^site)
   end
--- a/lib/beacon/layouts.ex
+++ b/lib/beacon/layouts.ex
@@ -21,10 +21,22 @@ defmodule Beacon.Layouts do
     Repo.all(Layout)
   end
 
-  def list_layouts_for_site(nil), do: []
-
-  def list_layouts_for_site(site) do
+  def list_layouts_for_site(site) when not is_nil(site) do
     Repo.all(from l in Layout, where: l.site == ^site)
+  end
+
+  @doc """
+  Returns a distinct list of sites associated with layouts.
+
+  ## Examples
+
+      iex> list_distinct_sites_from_layouts()
+      [:site_01, ...]
+
+  """
+  @spec list_distinct_sites_from_layouts :: list(Beacon.Type.Site.t())
+  def list_distinct_sites_from_layouts do
+    Repo.all(from l in Layout, distinct: true, select: l.site)
   end
 
   @doc """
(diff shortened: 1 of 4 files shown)


TheFirstAvenger:
@alexandrexaviersm Do we know why the site is nil in the first place? I would like to track the root cause of that first before tackling it from this angle.

alexandrexaviersm:
Yes, sorry for not putting more information:

This problem happens because we can render the `BeaconWeb.Admin.PageLive.FormComponent` with an empty `%Page{}` struct
https://github.com/BeaconCMS/beacon/blob/main/lib/beacon_web/live/admin/page_live/index.ex#L47
https://github.com/BeaconCMS/beacon/blob/main/lib/beacon_web/live/admin/page_live/index.html.heex#L29-L33

and inside the update callback from the `FormComponent` we get the `changeset` for this pages struct
```elixir
  def update(%{page: page} = assigns, socket) do
    changeset = Pages.change_page(page)
```
https://github.com/BeaconCMS/beacon/blob/main/lib/beacon_web/live/admin/page_live/form_component.ex#L10

which returns
```elixir
#Ecto.Changeset<
  action: nil,
  changes: %{},
  errors: [
    path: {"Not a string", []},
    site: {"can't be blank", [validation: :required]},
    template: {"can't be blank", [validation: :required]},
    layout_id: {"can't be blank", [validation: :required]},
    pending_template: {"can't be blank", [validation: :required]},
    pending_layout_id: {"can't be blank", [validation: :required]}
  ],
  data: #Beacon.Pages.Page<>,
  valid?: false
>
```

After that, we passed this `changeset` to the 'assign_layouts' function, which tries to get the field`: site`, but as it is an empty structure from the beginning, we get 'nil' as a result.
```elixir
  defp assign_layouts(%{assigns: %{changeset: changeset}} = socket) do
    layouts =
      changeset
      |> Ecto.Changeset.get_field(:site)
      |> Layouts.list_layouts_for_site()

    assign(socket, :site_layouts, layouts)
  end
```
https://github.com/BeaconCMS/beacon/blob/main/lib/beacon_web/live/admin/page_live/form_component.ex#L71-L78

leandrocp:
In short we need a site to create pages, but there's no way to create sites on admin at this moment. And that's related to https://elixir-lang.slack.com/archives/C02T04D147M/p1677088061416699 too (create and start sites at runtime dynamically).

leandrocp:
@alexandrexaviersm  but in this case, if you're using `mix dev` then it _should_ load because we're creating pages for site `dev`, which means either `dev.exs` has a bug or there's a bug when saving or loading the data. And that line `def list_layouts_for_site(nil), do: []` wouldn't be necessary.

alexandrexaviersm:
Thanks @leandrocp - I'm working on that!

leandrocp:
Thanks @alexandrexaviersm !

PR #146, opened 2023-02-27
Import app css

leandrocp:
We need to import the app css into the input source to evaluate tailwind functions and directives https://tailwindcss.com/docs/functions-and-directives.

Another option would be using PostCSS on the host application but I don't think asking users to do it is good UX if we can provide basic import out of the box.

### DIFF
--- a/test/beacon_web/live/page_live_test.exs
+++ b/test/beacon_web/live/page_live_test.exs
@@ -102,7 +102,6 @@ defmodule BeaconWeb.Live.PageLiveTest do
     test "a given path", %{conn: conn} do
       {:ok, _view, html} = live(conn, "/home")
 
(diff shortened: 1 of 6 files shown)


test/beacon_web/live/page_live_test.exs
```diff
@@ -102,7 +102,6 @@ defmodule BeaconWeb.Live.PageLiveTest do
     test "a given path", %{conn: conn} do
       {:ok, _view, html} = live(conn, "/home")
 
-      assert html =~ "body {cursor: zoom-in;}"
```
leandrocp, line 105: Removed because we're not running tailwind compiler on test env and that module already has specific tests.

TheFirstAvenger (approved):

PR #147, opened 2023-02-28
Add page.description

### LINKED ISSUE #145: page description
Along with title each page should have a description or summary field

leandrocp:
Close #145

### DIFF
--- a/lib/beacon/pages/page.ex
+++ b/lib/beacon/pages/page.ex
@@ -14,6 +14,7 @@ defmodule Beacon.Pages.Page do
   schema "beacon_pages" do
     field :site, Beacon.Type.Site
     field :title, :string
+    field :description, :string
     field :version, :integer, default: 1
     field :path, :string
     field :template, :string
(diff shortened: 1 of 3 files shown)


lib/beacon/pages/page.ex
```diff
@@ -14,6 +14,7 @@ defmodule Beacon.Pages.Page do
   schema "beacon_pages" do
     field :site, Beacon.Type.Site
     field :title, :string
+    field :description, :string
```
bcardarella, line 17: I think `string` is limited to 255 characters in Postgres, right? Wouldn't we want this to be a `text` instead?

leandrocp: `:string` here is the data representation. The actual hard limit is defined by the type used on the migration: https://github.com/BeaconCMS/beacon/blob/03a4cad48b8c21bd4ceafec43b92e295e452d7ac/priv/repo/migrations/20230228192357_add_description_to_pages.exs#L6
In this case I've used `text` for unlimited length and eventually we can add `validate_length` to enforce a soft limit on the app side.

bcardarella: oh sorry, yes missed that

leandrocp: No worries 👍🏻

PR #148, opened 2023-03-01
Update meta tags on page navigation

### LINKED ISSUE #133: Update head meta tags
https://github.com/BeaconCMS/beacon/pull/130#issuecomment-1419149872

leandrocp:
Close #133 

- [x] Push event with all meta tags
- [x] Remove current meta tags
- [x] Create the new meta tags received from event
- [x] Preserve (do not replace) csrf-token

https://user-images.githubusercontent.com/36407/222246834-853c39b8-4a57-464a-8d0a-6c72dc195117.mov


lib/beacon_web/components/layouts.ex
```diff
@@ -77,30 +77,28 @@ defmodule BeaconWeb.Layouts do
   See `Beacon.default_site_meta_tags/0` for a list of default meta tags
   that are included in all pages.
   """
-  def meta_tags(%{__dynamic_page_id__: _, __site__: _} = assigns) do
-    {:safe, meta_tags_unsafe(assigns)}
-  end
+  def render_meta_tags(assigns) do
```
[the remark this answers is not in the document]
leandrocp: Done thanks!

lib/beacon_web/components/layouts/runtime.html.heex
```diff
@@ -1,7 +1,8 @@
 <!DOCTYPE html>
 <html lang="en" phx-socket={live_socket_path(assigns)}>
   <head>
-    <%= meta_tags(assigns) %>
+    <meta name="csrf-token" content={get_csrf_token()} />
```
schrockwell, line 4: Is it okay to persist the CSRF token across page navigation? I am genuinely asking because I am not sure. The CSRF token is generated as part of the Plug lifecycle (disconnected mount), so I worry about it breaking if we don't update it during a live navigation.

schrockwell (approved):

lib/beacon_web/components/layouts/runtime.html.heex
leandrocp: This file is the actual root layout of the host application so it works the same as a regular Phoenix app. It gets this name to make it more explicit and to differentiate from the admin layout.

schrockwell: Ah! That makes sense. Thank you.

TheFirstAvenger (approved):

PR #153, opened 2023-03-06
AST rewrite

bcardarella:
3x performance improvement in tests against dockyard.com

### Tasks
- [x] Replace binaries with AST
- [x] Pre-compile template AST and store in ETS
- [x] Store routes in ETS - paths and metadata (page_id, layout_id, and modules)
- [x] Lookup ETS routes by path (including paths with dynamic segments)
- [x] Render templates dynamically from AST stored in ETS
- [x] `@beacon_path_params`
- [x] Propagate updates
- [x] Double check docs and generator (new apps)


bcardarella:
I've got to fix the test suite, and there are some additonal performance optimizations that can be gotten by reducing the number of functions generated.

lib/beacon/application.ex
```diff
@@ -14,6 +14,8 @@ defmodule Beacon.Application do
       Beacon.Repo
     ]
 
+    :ets.new(:beacon_pages, [:set, :named_table, :public, read_concurrency: true, write_concurrency: true])
```
bitwalker, line 17: Unless you plan to write to this table a lot, I would remove `write_concurrency` from this list. My assumption is that this is a table vastly dominated by reads. Without the option, you can of course still write to the table concurrently, it is just more expensive to do so.

lib/beacon/loader/db_loader.ex
```diff
@@ -18,10 +19,24 @@ defmodule Beacon.Loader.DBLoader do
   end
 
   def load_from_db(site) do
+    Logger.info("[DBLoader] Loading site #{site} from db")
+    Logger.info("[DBLoader] memory: #{inspect(:erlang.memory())}")
```
bitwalker, line 23: Might be better to emit telemetry events instead of logging this (or emit logs at `debug` level instead). I know with my various open source projects people have often requested control over what is logged and when, and `telemetry` really provides the best degree of control over this (as well as having other benefits).

lib/beacon/loader/module_loader.ex
```diff
@@ -1,20 +1,29 @@
 defmodule Beacon.Loader.ModuleLoader do
-  def load(module, code_string) do
+  def load(module, ast) do
     :code.delete(module)
```
bitwalker, line 3: I don't think you need (nor should use) either `:code.delete/1` or `:code.purge/1` here. If there are ongoing requests to the server, this will result in those requests failing badly if they happen to be requesting something which references the affected code. You should leave this behavior to the ERTS code loader itself, which will ensure that ongoing requests complete using the previous version of the code, and any new requests use only the new code.

bitwalker (approved):
I left a few suggestions, but I'll leave it up to you on whether you want to use them. Looks good otherwise!

lib/beacon/loader/module_loader.ex
bcardarella: this would only get kicked off once on a page CRUD event. But with the move towards storing compiled pages in ets the need for this at all should go away

lib/beacon/application.ex
leandrocp: It won't be written very often. I'm gonna remove that opt.

lib/beacon/loader/db_loader.ex
leandrocp: Yep that's planned. In this case it was a quick mechanism for debugging but it'll be removed before merging 👍🏻

TheFirstAvenger (approved):

bcardarella:
🎉

PR #168, opened 2023-03-17
allow beacon_admin to accept opts to configure live_session.

xn:
Admin routes for a beacon site.

  ## Examples

      defmodule MyAppWeb.Router do
        use Phoenix.Router
        use Beacon.Router

        scope "/", MyAppWeb do
          pipe_through :browser
          beacon_admin "/admin", on_mount: [SomeHook]
        end
      end

  ## Options

    * `:on_mount` (optional) , an optional list of `on_mount` hooks passed to `live_session`.
    This will allow for authenticated routes, among other uses.

### DIFF
--- a/lib/beacon/router.ex
+++ b/lib/beacon/router.ex
@@ -120,9 +120,26 @@ defmodule Beacon.Router do
   end
 
   @doc """
-  Admin routes.
+  Admin routes for a beacon site.
+
+  ## Examples
+
+      defmodule MyAppWeb.Router do
+        use Phoenix.Router
+        use Beacon.Router
+
+        scope "/", MyAppWeb do
+          pipe_through :browser
+          beacon_admin "/admin", on_mount: [SomeHook]
+        end
+      end
+
+  ## Options
+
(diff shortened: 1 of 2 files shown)


lib/beacon/router.ex
```diff
@@ -153,6 +172,22 @@ defmodule Beacon.Router do
     end
   end
 
+  @doc false
+  def __admin_session_opts__(opts) do
+    if Keyword.has_key?(opts, :root_layout) do
```
leandrocp, line 177: Let's make `:layout` reserved as well.

leandrocp (changes requested):
Nice one. Just one comment to make `:layout` reserved just like `:root_layout` and you'll need to import `beacon_admin/2` here https://github.com/BeaconCMS/beacon/blob/f4072b2e62fc0757df6186445e1b06f16df2951a/lib/beacon/router.ex#L21

lib/beacon/router.ex
xn: I missed that. Thanks!

leandrocp (approved):
🎉

PR #173, opened 2023-03-21
Smar reload pages

### LINKED ISSUE #76: Smart reload on publish page
Currently, when [publishing a single page](https://github.com/BeaconCMS/beacon/blob/main/lib/beacon/pages.ex#L101), we call `load_from_db` which [reloads](https://github.com/BeaconCMS/beacon/blob/d0a75a593257103d267d7a0e3b02b5fccd1956e2/lib/beacon/loader/db_loader.ex#L12-L15) every page, component, layout, and stylesheet. We should instead call a new function that reloads the single page and the layout it uses (since the layout needs to regenerate the css based on the new page).

leandrocp:
Close #76 

- [ ] Merge DBLoader, Loader.Server, and ModuleLoader into `Beacon.Loader` so every request to reload resources goes through the Loader process to avoid overflowing the system if users submit too many requests, and having a single module is easier to reason about the workflow (less indirections).
- [ ] Break down layout and page modules into 1 module per resource so it can be reloaded individually.
- [ ] Add public api functions to load all sites, load a single site, or load a single page.
- [ ] Small name changes to make things more explicit.
- [ ] Group errors into a single module.
- [ ] Start `Loader` in specific tests that need it.


leandrocp:
There's 1 task pending that won't impact the others so it's ready for review.

TheFirstAvenger (approved):

PR #174, opened 2023-03-22
Authorization spike

xn:
Did a quick walk through with @leandrocp 

currently sorting out the shape of `operation_context`
need to add some guard clauses around ux items
need to sort out many sites vs one admin config wise.
need to add query filtering skeleton


lib/beacon/authorization.ex
```diff
@@ -0,0 +1,104 @@
...
+      that matches these args.\
+      """
+
+      reraise __MODULE__.Error, [message: error_message], __STACKTRACE__
+
+    error ->
+      reraise error, __STACKTRACE__
+  end
+
+  defp get_authorization_source() do
```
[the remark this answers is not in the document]
xn: I can work with that. Thanks!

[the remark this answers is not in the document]
xn: if we call it `:admin_defaults`, then we leave the door open for per site admin configuration only as needed.

lib/beacon/authorization/behavior.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Beacon.Authorization.Behaviour do
```
leandrocp, line 1: Hey @xn I was thinking about that callback for the pages query that we discussed yesterday. What if we don't have provide such callback for now? The `authorized?` callback will limit who can edit or delete each page, which I believe is enough for now, and making the list of pages public for all users is okay. I feel that adding the query callback complexity isn't worth the value it will bring, unless I'm missing something. Wdyt?

schrockwell, line 1: I feel like putting the word "Behaviour" in the module name is not generally accepted. Here are some other names for brainstorming.

```
Beacon.Authorization
Beacon.Authorization.Policy
Beacon.Authorization.Rules
```

lib/beacon/authorization/behavior.ex
```diff
@@ -0,0 +1,20 @@
...
+
+  """
+
+  # returns requestor_context
+  @callback get_requestor_context(payload :: any()) :: any()
+
+  # site, requestor_context, operation_context (needs type)
+  @callback authorized?(Beacon.Type.Site.t(), requestor_context :: any(), operation_context :: map()) :: boolean()
+  # requestor_context, operation_context (needs type)
+  @callback authorized?(requestor_context :: any(), operation_context :: map()) :: boolean()
```
schrockwell, line 15: I don't love having two functions that differ just by arity. Can we either (a) combine these into `authorized?/3` and have a "reserved" atom for the first argument, or (b) rename `authorized?/2`?

lib/beacon/authorization/behavior.ex
leandrocp: I believe @xn was following the existing pattern where we have `Beacon.DataSource.Behaviour` but I agree with this change, it makes more sense. We can even rename `Beacon.DataSource.Behaviour` in a following PR, we're okay with making breaking changes while it's still pre-release.

xn: We definitely don't have to do so now.

xn: I feel you. I was following the pattern set up with `Beacon.DataSource.Behaviour`

If we want to break that pattern, I'd say we should do it in a separate PR and unify the both of them.

lib/beacon/router.ex
```diff
@@ -190,12 +190,33 @@ defmodule Beacon.Router do
       """
     end
 
+    on_mounts = get_on_mount_list(Keyword.get(opts, :on_mount))
```
schrockwell, line 193:
```suggestion
    on_mounts = get_on_mount_list(Keyword.get(opts, :on_mount, []))
```

and then you can remove both these clauses:

```elixir
  defp get_on_mount_list(nil), do: [BeaconWeb.Hooks.AssignAuthorizationContext]
  defp get_on_mount_list([]), do: [BeaconWeb.Hooks.AssignAuthorizationContext]
```

lib/beacon/authorization/behavior.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Beacon.Authorization.Behaviour do
+  @moduledoc """
+    Provides hooks into Beacon Authorization
+
+    ## Examples
+
+  """
+
+  # returns requestor_context
+  @callback get_requestor_context(payload :: any()) :: any()
```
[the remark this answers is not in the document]
xn: There are two contexts to add together, one coming from the authenticated party and the other from the beacon.

I loathe calling specific things merely context.

lib/beacon/config.ex
```diff
@@ -12,6 +12,9 @@ defmodule Beacon.Config do
     * `:data_source` (optional) `t:data_source/0` - a module that implements `Beacon.DataSource.Behaviour`,
       used to provide assigns to your site pages.
 
+    * `:authorization_source` (optional) `t:authorization_source/0` - a module that implements `Beacon.Authorization.Behaviour`,
+      used to provide authorization rules for the admin backend.
```
leandrocp, line 16:
```suggestion
      used to provide authorization rules for the admin backend. Defaults to `Beacon.Authorization.DefaultPolicy` and note this config can't be `nil`.
```

Feel free to change wording here but we have to make it explicit this config can't be `nil` because admin depends on it. It's optional because we provide a default policy but it has to be valid.

lib/beacon_web/hooks/assign_agent.ex
```diff
@@ -0,0 +1,14 @@
+defmodule BeaconWeb.Hooks.AssignAgent do
+  import Phoenix.Component
+
+  alias Beacon.Authorization
+
+  def on_mount(:default, _params, session, socket) do
+    authorization_bootstrap = Map.get(session, "authorization_bootstrap", nil)
```
leandrocp, line 7: I'm assuming you'll be passing `authorization_bootstrap` from the client, right? Which means this value is a public contract and users should know about it. Can we discuss about it when you have some time?

leandrocp (approved):
@xn the first iteration looks good! Feel free to merge it.

@TheFirstAvenger for context, we agreed to merge it and move forward with this first version to experiment how it feels to use it in a real app and I'll push a PR making some small changes.

PR #184, opened 2023-04-10
Introduce the Lifecycle API - inject custom user-defined steps into beacon

leandrocp:
Allow consuming apps (host apps) to extend and integrate with Beacon internals by defining custom steps into the following stages: create page, publish page, load template, and render template.

- [x] Store format on page metadata ets
- [x] Stage: Page load
- [x] Stage: Page render
- [x] Stage: Page create
- [x] Stage: Page publish
- [x] Template format heex
- [x] Template format markdown
- [x] Separate /admin assets into its own domain - that is necessary to render an instance of monaco editor without impacting the sites assets, the initial work is done but I'll leave the actual implementation to a following PR because this one is getting too long, that's why you'll see some leftover JS code.


lib/beacon/config.ex
```diff
@@ -63,7 +117,35 @@ defmodule Beacon.Config do
             css_compiler: Beacon.TailwindCompiler,
             tailwind_config: Path.join(Application.app_dir(:beacon, "priv"), "tailwind.config.js.eex"),
             live_socket_path: "/live",
-            safe_code_check: false
+            # TODO: change safe_code_check to true when it's ready to parse complex codes
+            safe_code_check: false,
+            template_formats: [
```
leandrocp, line 122: This API can be considered done but I'm open to suggestions. Initially I thought about a API to allow users adding or removing steps from HEEx or Markdown formats so they could add custom steps to transform them. But that proved to be more complex than just letting the user define the entire lifecycle (all steps to `load` and `render` a template in a given format) with the possibility of reusing steps like I've done for Markdown that's reusing HEEx steps.

Note that also opens the possibility to implement other engines quite easily, for eg Liquid which is a simple and safe template language. Even thought that's not the objective, I believe it's nice to have since there's no need to write extra code to support that.

lib/beacon/config.ex
```diff
@@ -2,81 +2,247 @@ defmodule Beacon.Config do
...
+             [
+               assigns: fn template, %{assigns: assigns} -> MyEngine.parse_to_html(template, assigns) end,
+               compile: &Beacon.Template.HEEx.compile/2
+               eval: &Beacon.Template.HEEx.eval_ast/2
+             ]}
+          ],
+          publish_page: [
+            notify_admin: fn page -> {:cont, MyApp.send_email(page) end
+          ]
+        ]
```
leandrocp, line 213: This API can be considered done but I'm open to suggestions. Initially I thought about a API to allow users adding or removing steps from HEEx or Markdown formats so they could add custom steps to transform them. But that proved to be more complex than just letting the user define the entire lifecycle (all steps to load and render a template in a given format) with the possibility of reusing steps like I've done for Markdown that's reusing HEEx steps.

And the reasons to leverage the config keyword instead of functions are to keep all config in the same place, it's easier to inspect and debug and the most important is that we need the `:load_template` steps as early as possible to boot the app so it has to be available when the site is registered.

TheFirstAvenger (approved):

PR #237, opened 2023-05-04
Lifecycle API - add explicit callbacks and lifecycle struct and move the handling of stages to its own modules

xn:
We're about to add multiple lifecycles and the current one module approach would hold more than one concern, so we're breaking it up.

The `Lifecycle` module now has two main functions:
`fetch_steps!` and `execute_steps`

All other code has been moved to child modules.


xn:
Looking at this, I want to change the parameter order of `execute_steps` so we can easily pipe these functions.

leandrocp:
> Looking at this, I want to change the parameter order of `execute_steps` so we can easily pipe these functions.

Go ahead

PR #239, opened 2023-05-15
Asset Configuration

xn:
Adds config for assets allowed, backends and validations.

- Whitelist specific mime types. `["image/gif", "image/png", "image.jpeg"]
- Assign multiple backends to mime type or catchall: `"image/jpeg"` or `"image/*"`


lib/beacon/config.ex
```diff
@@ -298,16 +314,59 @@ defmodule Beacon.Config do
...
     struct!(__MODULE__, opts)
   end
 
   @spec fetch!(Beacon.Types.Site.t()) :: t()
-  def fetch!(site) do
+  def fetch!(site) when is_atom(site) do
     Registry.config!(site)
   end
+
+  @spec config_for_mime_type(Beacon.Config.t(), MIME.t()) :: term()
```
leandrocp, line 354: you could move `Enum.into(%{})` into `config_for_mime_type/2`
```suggestion
  @spec config_for_mime_type(Beacon.Config.t(), MIME.t()) :: map()
```

lib/beacon/config.ex
```diff
@@ -48,6 +48,14 @@ defmodule Beacon.Config do
   """
   @type template_formats :: [{format :: atom(), description :: String.t()}]
 
+  @type asset :: [{MIME.t(), mime_config()}]
```
leandrocp, line 51: I don't think `MIME.t/0` exists, it's just a `String.t/0` :)
but we can have our own custom type

lib/beacon/admin/media_library/file_metadata.ex
```diff
@@ -0,0 +1,36 @@
...
+
+  def new(site, path, name, mime_type, size) do
+    config = Beacon.Config.fetch!(site)
+
+    asset_config =
+      config
+      |> Beacon.Config.config_for_mime_type(mime_type)
+      |> Enum.into(%{})
+
+    %__MODULE__{
```
leandrocp, line 26: I see some inconsistency on the "type" key:
- config.asset_formats
- metadata.allowed_types
- metadata.mime_type
- asset.file_type

The term "MIME type" was updated to "Media Type" as noted at https://www.iana.org/assignments/media-types/media-types.xhtml so we have 3 acceptable suggestions:
- `media_type` - technically the correct name
- `mime_type` - still popular
- `type`

I'm more inclined to suggest calling it just `type` because `image/*` isn't a registered and valid media type and IANA calls the main category`image` as "type", see https://www.iana.org/assignments/media-types/image/webp - that would look like:
- config.asset_types
- metadata.allowed_types
- metadata.type
- asset.type

Wdyt?

leandrocp (commented):
Hey @xn some comments before we can move forward. Thanks

lib/beacon/admin/media_library/provider/repo.ex
```diff
@@ -0,0 +1,16 @@
+defmodule Beacon.Admin.MediaLibrary.Provider.Repo do
```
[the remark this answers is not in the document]
xn: Backend.Repo is just fine.

lib/beacon/config.ex
```diff
@@ -298,16 +314,59 @@ defmodule Beacon.Config do
...
   end
 
   @spec fetch!(Beacon.Types.Site.t()) :: t()
-  def fetch!(site) do
+  def fetch!(site) when is_atom(site) do
     Registry.config!(site)
   end
+
+  @spec config_for_mime_type(Beacon.Config.t(), MIME.t()) :: term()
+  def config_for_mime_type(beacon_config, mime_type) do
```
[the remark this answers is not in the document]
xn: Good catch

lib/beacon/admin/media_library/file_metadata.ex
xn: You are correct on multiple accounts here.
- `*` is not a valid mime sub type
- naming variance.

`image/*` is  a valid `content type`, however.

I'm comfortable with your name changes. Does `content_type` hold anything for you? It can point the way on what is expected there.

leandrocp: > image/* is a valid content type, however

that's a good point, I'm okay with using `content_type`

xn: I was also glomming onto what Plug.Conn was doing:

https://hexdocs.pm/plug/Plug.Conn.Utils.html#media_type/1

lib/beacon/config.ex
xn: You're right. I meant to ask if you wanted anything special there and that was a place keeper.

lib/beacon/config.ex
xn: I guess the question here is, this function currently returns data of a type encoded by this module and do we want to return something else? I think when I originally wrote this, I hadn't created the `mime_config` type.

So it should be ```@spec config_for_mime_type(Beacon.Config.t(), MIME.t()) :: mime_config()```
unless you really want to return a map there.

dev.exs
```diff
@@ -283,19 +283,27 @@ seeds = fn ->
...
-    "image/png"
-  )
-
-  Beacon.Admin.MediaLibrary.upload(
-    "dev",
-    Path.join(:code.priv_dir(:beacon), "assets/dockyard.png"),
-    "dockyard_2.jpg",
-    "image/jpg"
-  )
+  metadata =
```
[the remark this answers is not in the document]
leandrocp: I see the benefits for testing and I agree that 5+ args is too much (it has 4 now). I'm just worried that we're exposing `FileMetadata` internals that should be private. One could write `upload(%FileMetadata{config: nil})` and break the pipeline, which I see is possible since the spec would look like `@spec upload(FileMetadata.t)`, right?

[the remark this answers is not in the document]
leandrocp: Can you please add a `TODO: https://github.com/BeaconCMS/beacon/pull/239#discussion_r1194160478` close to `def new`? To serve as a reminder to revisit it before v0.1 so we can move forward with this PR.

lib/beacon/config.ex
leandrocp: `String.t()` or a custom `ContentType.t()` works for me, whatever you prefer 👍🏻

lib/beacon/admin/media_library/file_metadata.ex
leandrocp: Sorry for the confusion but giving a second thought, I think it's better to call it `media_type` for the following reasons:
- `type` is too generic
- `media_type` is technically correct
- `content_type` serves to another purpose
- having `image/*` as the media type is fine since content type and Plug accepts it 

Sounds good?

lib/beacon/config.ex
leandrocp: > unless you really want to return a map there.

Nope. It was just to avoid forcing the caller to transform the returned data.

lib/beacon/admin/media_library/file_metadata.ex
xn: works for me!

xn:
Thanks for the great feedback, @leandrocp !

dev.exs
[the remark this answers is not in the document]
xn: Will do!

A developer could write `upload(%FileMetadata{config: nil})` without a catchall, but it would have to go through a PR process.

If you simply prefer to have the FileMetadata constructed inside the function, that is fine and I'm happy to change it. 

Doing that though means either upload needs to accept a file size, or we need to calculate it somewhere.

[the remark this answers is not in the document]
leandrocp: Let's keep it as it's and see how it plays as we work on Admin

leandrocp (approved):
Nice one

PR #242, opened 2023-05-18
convert image files to webp by default

xn:
This gets the idea going. When we add thumbnails, I'm sure this will turn into a Lifecycle.

Right now, we convert all images to webp. This can be overridden by using your own processor.

Minimum version of elixir supported is bumped to 1.13.0


TheFirstAvenger (approved):

xn:
@leandrocp Which versions of Elixir are we targeting? Image is failing to compile on 1.12.0

leandrocp:
> @leandrocp Which versions of Elixir are we targeting? Image is failing to compile on 1.12.0

1.12 but we can bump min that supports Image

test/support/fixtures/image.gif
leandrocp, line 1: Not a blocker but can we reduce the size of those support images?

lib/beacon/config.ex
```diff
@@ -63,7 +63,10 @@ defmodule Beacon.Config do
   """
   @type media_type_config :: [
           {:backends, list(backend :: module())},
-          {:validations, list(validation_fun :: (Ecto.Changeset.t(), Beacon.Admin.MediaLibrary.FileMetadata.t() -> Ecto.Changeset.t()))}
+          {:validations, list(validation_fun :: (Ecto.Changeset.t(), Beacon.Admin.MediaLibrary.UploadMetadata.t() -> Ecto.Changeset.t()))},
+          {:processor,
+           {prossesor_fun ::
```
leandrocp, line 68: If I'm not wrong, I see this fn called with 1 arg. Is this spec correct?

mix.exs
```diff
@@ -43,6 +43,7 @@ defmodule Beacon.MixProject do
       {:floki, ">= 0.30.0", only: :test},
       {:gettext, "~> 0.20"},
       {:heroicons, "~> 0.5"},
+      {:image, "~> 0.31.1"},
```
leandrocp, line 46:
```suggestion
      {:image, "~> 0.31"},
```

As a lib we can't be too restrictive :)

leandrocp (approved):

lib/beacon/config.ex
xn: You are correct. I hadn't updated that to my final implementation. Great catch!

test/support/fixtures/image.gif
xn: is 100k fine or do you want tiny?

leandrocp: 100k is fine 👍🏻

PR #249, opened 2023-06-02
Add `:default_meta_tags` config option

APB9785:
### Asana task
[Improve meta tag functionality](https://app.asana.com/0/1203282419046399/1204726099802197/f)

### Goal
We want sites to be able to configure default meta tags which are automatically added to new pages

### Implementation
Adds an optional `:default_meta_tags` to `Beacon.Config`, which is just a list of meta tags (maps).  Using meta tag maps allow the most customization e.g. setting default content values, name vs. property.

### DIFF
--- a/lib/beacon_web/live/admin/page_live/form_component.ex
+++ b/lib/beacon_web/live/admin/page_live/form_component.ex
@@ -39,7 +39,9 @@ defmodule BeaconWeb.Admin.PageLive.FormComponent do
   end
 
   defp save_page(socket, :new, page_params) do
-    case Pages.create_page(page_params) do
+    parsed_params = Map.new(page_params, fn {key, value} -> {String.to_existing_atom(key), value} end)
+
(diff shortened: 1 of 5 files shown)


lib/beacon_web/live/admin/page_live/form_component.ex
```diff
@@ -39,7 +39,9 @@ defmodule BeaconWeb.Admin.PageLive.FormComponent do
   end
 
   defp save_page(socket, :new, page_params) do
-    case Pages.create_page(page_params) do
+    parsed_params = Map.new(page_params, fn {key, value} -> {String.to_existing_atom(key), value} end)
+
+    case Pages.create_page(parsed_params) do
```
APB9785, line 44: Elsewhere we create pages with atom-key params, so here we also convert to atom keys

leandrocp (approved):
Thanks!

PR #252, opened 2023-06-09
Update ecto from 3.9.4 to 3.9.5 to work on OTP26

cibernox:
I found this problem setting things up on the latest OPT. Googling, seems that ecto released a simple patch version with a fix and no other breaking changes, so this should be safe.


cibernox:
@AZholtkevych to clarify, this is probably NOT all that that's needed to support OTP26 and Elixir 1.15, but it brings us closer.

PR #259, opened 2023-06-19
Add ui-builder API 

### LINKED ISSUE #2: Editing and creating new data sources
Currently all data sources require a re-deployment to make changes. These should be editable at runtime. At the very least setting custom data source scoping per page path.

cibernox:
Adds the basic API needed for the UI builder to render components and add them to the page.

Closes #2 

There will be some refinements, particularly around reordering components, which right now is not possible. Possibly copy-pasting components would be extremely handy in practice. Or being able to drag into the sidebar (sometimes is tricky to aim to the correct component, is would be easier to select it first and then drag another component inside using the sidebar)


lib/beacon/component_categories.ex
```diff
@@ -0,0 +1,49 @@
...
+  Returns the list of component categories.
+
+  ## Examples
+
+      iex> list_component_categories()
+      [%ComponentCategories{}, ...]
+
+  """
+  def list_component_categories do
+    ComponentCategory |> Repo.all()
```
APB9785, line 21:
```suggestion
    Repo.all(ComponentCategory)
```
No need to use the pipe operator if there is only one

lib/beacon/component_definitions.ex
```diff
@@ -0,0 +1,65 @@
...
+  Returns the list of component definitions.
+
+  ## Examples
+
+      iex> list_component_definitions()
+      [%ComponentDefinitions{}, ...]
+
+  """
+  def list_component_definitions do
+    ComponentDefinition |> Repo.all()
```
APB9785, line 21:
```suggestion
    Repo.all(ComponentDefinition)
```

lib/beacon/component_definitions.ex
```diff
@@ -0,0 +1,65 @@
...
+  ## Examples
+
+      iex> get_component_definition!(123)
+      %ComponentDefinition{}
+
+      iex> get_component_definition!(456)
+      ** (Ecto.NoResultsError)
+
+  """
+  def get_component_definition!(id, preloads \\ []), do: ComponentDefinition |> Repo.get!(id) |> Repo.preload(preloads)
```
APB9785, line 38:
```suggestion
  def get_component_definition!(id, preloads \\ []) do
    ComponentDefinition
    |> Repo.get!(id)
    |> Repo.preload(preloads)
  end
```
`do/end` block is preferred for pipelines

lib/beacon/component_instances.ex
```diff
@@ -0,0 +1,71 @@
...
+  Returns the list of component instances.
+
+  ## Examples
+
+      iex> list_component_instances()
+      [%ComponentInstance{}, ...]
+
+  """
+  def list_component_instances do
+    ComponentInstance |> Repo.all()
```
APB9785, line 21:
```suggestion
    Repo.all(ComponentInstance)
```

lib/beacon/component_instances.ex
```diff
@@ -0,0 +1,71 @@
...
+  ## Examples
+
+      iex> get_component_instance!(123)
+      %ComponentInstance{}
+
+      iex> get_component_instance!(456)
+      ** (Ecto.NoResultsError)
+
+  """
+  def get_component_instance!(id, preloads \\ []), do: ComponentInstance |> Repo.get!(id) |> Repo.preload(preloads)
```
APB9785, line 38:
```suggestion
  def get_component_instance!(id, preloads \\ []) do
    ComponentInstance
    |> Repo.get!(id)
    |> Repo.preload(preloads)
  end
```

lib/beacon/content.ex
```diff
@@ -19,6 +21,7 @@ defmodule Beacon.Content do
   alias Beacon.Repo
   alias Beacon.Types.Site
   alias Ecto.Changeset
+  import Ecto.Query
```
APB9785, line 24: Ideally we want to keep module directives like this organized in blocks of descending impact, with an empty line in-between. i.e.

```elixir
use FooModule

import BarModule

alias BazModule
alias BongModule

require SomethingElse
```

Since `import` has a greater impact on this module than `alias`, it should go above

lib/beacon_web/controllers/api/component_controller.ex
```diff
@@ -0,0 +1,67 @@
+defmodule BeaconWeb.API.ComponentController do
+  use BeaconWeb, :controller
+  require Logger
+  alias Beacon.ComponentCategories
+  alias Beacon.ComponentDefinitions
+  alias Beacon.ComponentInstances
+  alias Ecto.UUID
```
APB9785, line 7:
```suggestion
  use BeaconWeb, :controller
  
  alias Beacon.ComponentCategories
  alias Beacon.ComponentDefinitions
  alias Beacon.ComponentInstances
  alias Ecto.UUID
  
  require Logger
```

lib/beacon_web/controllers/api/component_controller.ex
```diff
@@ -0,0 +1,67 @@
...
+    definition = ComponentDefinitions.get_component_definition!(component_definition_id)
+    component_data = build_component(definition.blueprint)
+    render(conn, :show, component: %{id: UUID.generate(), data: component_data})
+  end
+
+  @spec update(Plug.Conn.t(), map) :: Plug.Conn.t()
+  def update(conn, %{"id" => id} = map) do
+    data = Map.delete(map, "id")
+
+    case ComponentInstances.get_component_instance!(id) |> ComponentInstances.update_component_instance_data(data) do
```
APB9785, line 46:
```suggestion
    instance = ComponentInstances.get_component_instance!(id)
    
    case ComponentInstances.update_component_instance_data(instance, data) do
```
Generally pipes should not be used inside a `case` head.  Instead, do all the work beforehand except for the last line, assign that result to a variable, and then put the last line inside the `case` head

lib/beacon_web/controllers/api/component_controller.ex
```diff
@@ -0,0 +1,67 @@
...
+      {:error, changeset} ->
+        json(conn, changeset.errors)
+    end
+  end
+
+  defp build_component(entry) when is_binary(entry), do: entry
+  defp build_component(%{"tag" => "raw", "attributes" => _, "content" => content}), do: content
+
+  defp build_component(%{"tag" => tag, "attributes" => attributes, "content" => content}) do
+    attributes = attributes |> Map.put("id", UUID.generate())
```
APB9785, line 59:
```suggestion
    attributes =  Map.put(attributes, "id", UUID.generate())
```

lib/beacon_web/controllers/api/component_controller.ex
```diff
@@ -0,0 +1,67 @@
...
+    content = content |> Enum.map(&build_component(&1))
```
APB9785, line 60:
```suggestion
    content = Enum.map(content, &build_component(&1))
```

lib/beacon_web/controllers/api/component_json.ex
```diff
@@ -0,0 +1,85 @@
...
+    <#{tag}#{render_attrs(attributes)}>
+      #{content |> Enum.map_join(&render_node(&1))}
+    </#{tag}>
+    """
+  end
+
+  defp render_attrs(attributes) when attributes == %{}, do: ""
+
+  defp render_attrs(attributes) do
+    str = attributes |> Enum.map_join(" ", fn {key, val} -> render_attr(key, val) end)
```
APB9785, line 75:
```suggestion
    str = Enum.map_join(attributes, " ", fn {key, val} -> render_attr(key, val) end)
```

lib/beacon_web/controllers/api/component_json.ex
```diff
@@ -0,0 +1,85 @@
...
+    " " <> str
+  end
+
+  defp render_attr(key, val) when is_list(val), do: "#{key}=\"#{val |> Enum.join(" ")}\""
```
APB9785, line 79:
```suggestion
  defp render_attr(key, val) when is_list(val), do: "#{key}=\"#{Enum.join(val, " ")}\""
```

lib/beacon_web/controllers/api/page_controller.ex
```diff
@@ -1,10 +1,15 @@
 defmodule BeaconWeb.API.PageController do
   use BeaconWeb, :controller
-
   alias Beacon.Content
+  alias Beacon.Repo
 
   action_fallback BeaconWeb.API.FallbackController
 
+  def index(conn, _params) do
+    pages = Content.list_pages(:dev) |> Repo.preload(:components)
```
APB9785, line 9:
```suggestion
    pages = 
      :dev
      |> Content.list_pages()
      |> Repo.preload(:components)
```
When possible, pipelines should always start with a literal, as opposed to a function call

lib/beacon_web/controllers/api/page_json.ex
```diff
@@ -22,7 +22,10 @@ defmodule BeaconWeb.API.PageJSON do
       path: page.path,
       site: page.site,
       template: page.template,
-      format: page.format
+      format: page.format,
+      components:
+        page.components
+        |> Enum.map(&BeaconWeb.API.ComponentJSON.show(%{component: &1}))
```
APB9785, line 28:
```suggestion
      components: Enum.map(page.components, &BeaconWeb.API.ComponentJSON.show(%{component: &1}))
```

dev.exs
```diff
@@ -360,6 +1338,203 @@ seeds = fn ->
...
+    name: "Button",
+    thumbnail: "/component_thumbnails/button.jpg",
+    blueprint: ComponentDefinitionHelpers.create_definition("button", %{}, ["I'm a sample button"]),
+    component_category_id: basics.id
+  })
+
+  Beacon.ComponentDefinitions.create_component_definition!(%{
+    name: "Link",
+    thumbnail: "/component_thumbnails/link.jpg",
+    blueprint: ComponentDefinitionHelpers.create_definition("a", %{href: "#"}, ["I'm a sample link"]),
```
leandrocp, line 1521: Since we are aiming to parse HEEx templates we don't need to store the intermediary data structure anymore, instead this should be a HEEx template string:

```suggestion
    blueprint: """
    <a href="#">I'm a sample link</a>
    """
```

We'll need to implement the HEEx parser to generate the JSON that Page Builder expects.

lib/beacon/content/component_definition.ex
```diff
@@ -0,0 +1,28 @@
...
+
+  @type t :: %__MODULE__{}
+
+  @primary_key {:id, :binary_id, autogenerate: true}
+  @foreign_key_type :binary_id
+  schema "beacon_component_definitions" do
+    field :name, :string
+    field :thumbnail, :string
+    field :blueprint, :map
+    belongs_to :component_category, ComponentCategory
```
leandrocp, line 15: We don't need `Beacon.ComponentCategories` **if** the only attr is `name` and we won't expose that to users create custom categories. It can be a [enum](https://hexdocs.pm/ecto/Ecto.Enum.html) instead:

```suggestion
    field : component_category, Ecto.Enum, values: [:header, :sign_in, :sign_up, :stats, :footer, :basic]
```

lib/beacon/component_definitions.ex
cibernox: Gotcha.

lib/beacon/component_instances.ex
```diff
@@ -0,0 +1,75 @@
+defmodule Beacon.ComponentInstances do
```
leandrocp, line 1: Hey @cibernox can you please describe the purpose of `ComponentInstance`? I'm not following why this is necessary, I may be missing something.

dev.exs
leandrocp: HEEx support to be added in a following PR based on https://github.com/BeaconCMS/beacon/issues/293

leandrocp (commented):
Hey @cibernox this is looking great. 

I'm trying to understand if we really need `ComponentCategory` and `ComponentInstance`, maybe I'm missing something but are these schemas necessary?

lib/beacon/blueprint_converter.ex
```diff
@@ -0,0 +1,39 @@
...
+    case Floki.parse_document(html_string) do
+      { :ok, data } ->
+        reshape(data)
+      { _, error} ->
+        raise("Error parsing blueprint")
+    end
+  end
+
+  defp convert_attr({"class", value}) do
+    { "class", value |> String.split(" ") }
```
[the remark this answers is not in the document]
cibernox: classes specifically can only be `String[]`. Most (all?) other attributes are either strings or true/false (those are technically props)

lib/beacon/blueprint_converter.ex
```diff
@@ -0,0 +1,39 @@
...
+  end
+  defp convert_attr({"href", value}), do: {"href", value}
+  defp convert_attr({"src", value}), do: {"src", value}
+  defp convert_attr(_), do: nil
```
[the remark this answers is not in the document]
cibernox: i started with a whitelist approach, because the UI only supports those, but I think I can remove this now.

lib/beacon/component_instances.ex
cibernox: Since I wanted to leave the `Beacon.component` table alone, I created this one. From a ComponentDefinition, which is like a blueprint, we create ComponentInstances which is the ones we assign to a specific page.

Do you reckon I should use the `Beacon.Content.Component` type instead of created a new one?

lib/beacon/blueprint_converter.ex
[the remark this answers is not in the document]
cibernox: Removed now. All attributes are supported.

lib/beacon/content/component_definition.ex
cibernox: Users will not create their own categories, I like the idea of removing one table

dev.exs
```diff
@@ -142,14 +149,6 @@ seeds = fn ->
     content: "body {cursor: zoom-in;}"
   })
 
-  Beacon.Content.create_component!(%{
-    site: "dev",
-    name: "sample_component",
-    body: """
-    <%= @val %>
-    """
-  })
```
leandrocp, line 151: Please revert this change, this component is used on the home page for local development.

lib/beacon/loader.ex
```diff
@@ -105,9 +114,11 @@ defmodule Beacon.Loader do
 
   # TODO: replace my_component in favor of https://github.com/BeaconCMS/beacon/issues/84
   defp load_components(site) do
+    components = Beacon.Content.list_components(site)
+
     Beacon.Loader.ComponentModuleLoader.load_components(
       site,
-      Beacon.Content.list_components(site)
```
leandrocp, line 110: You can revert this change too

priv/templates/install/seeds.exs
```diff
@@ -100,3 +100,155 @@ Content.publish_layout(layout)
...
+Beacon.ComponentDefinitions.create_component_definition!(%{
+  name: "Paragraph",
+  thumbnail: "/component_thumbnails/paragraph.jpg",
+  component_category_id: basics.id
+})
+Beacon.ComponentDefinitions.create_component_definition!(%{
+  name: "Aside",
+  thumbnail: "/component_thumbnails/aside.jpg",
+  component_category_id: basics.id
+})
```
leandrocp, line 254: I don't think think we need this block of code anymore, right?

```suggestion
```

lib/beacon_web/controllers/api/component_json.ex
```diff
@@ -0,0 +1,40 @@
+defmodule BeaconWeb.API.ComponentJSON do
+  alias Beacon.Content.Component
+
+  def index(%{component_definitions: definitions}) do
+    %{
+      menuCategories: [
```
leandrocp, line 6: Usually Phoenix APIs use the snake case pattern since that's the pattern used in the code base on the backend side. Is it okay if we change it to?

```suggestion
      menu_categories: [
```

leandrocp (changes requested):
Nice one @cibernox we're close to merge this one. Just left a couple of suggestions.

priv/templates/install/seeds.exs
cibernox: Removed

dev.exs
cibernox: Done

cibernox:
@leandrocp All feedback addressed

lib/beacon_web/controllers/api/component_json.ex
cibernox: Lets leave it camel cased and try to use a plug to convert between camelCase and snake_case transparently.

leandrocp (approved):
🎉

PR #267, opened 2023-06-27
add drag and drop, cancel upload, show preview, etc

xn:
Screenshot 2023-06-27 at 1 53 33 PM


leandrocp (approved):
Looks good!

Just remember to revert priv/static/beacon_admin.css and priv/static/beacon_admin.js - we still need to separate the assets generated by `mix dev` from the actual release assets.

PR #269, opened 2023-06-28
Add etag support for assets

xn:
Grabbed code from:
https://github.com/hexpm/hexpm/

Allows us to etag any future schema and is aware of associations.
We could apply this page controllers, etc.

### DIFF
--- a/lib/beacon/media_library/asset.ex
+++ b/lib/beacon/media_library/asset.ex
@@ -2,6 +2,7 @@ defmodule Beacon.MediaLibrary.Asset do
   @moduledoc false
 
   use Ecto.Schema
+  @derive BeaconWeb.Cache.Stale
 
   @primary_key {:id, :binary_id, autogenerate: true}
   @foreign_key_type :binary_id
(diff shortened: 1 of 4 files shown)


xn:
@bcardarella I thought it was MIT. I'll have to can this. We'd have to license this under the Apache 2.0 and document changes.

bcardarella:
@xn let's leave this open, I've asked permission if we can use this code outside the license requirements. Let's see if that can be done before we abandon

leandrocp:
AFAIK you can reuse and redistribute that code as long as you keep a header mentioning the original source, copyright, and license. Something like:

```elixir
# Copied and modified from the original work available at https://github.com/hexpm/hexpm/blob/d5b4e3864e219dd8b2cca34570dbe60257bcc547/lib/hexpm_web/controllers/controller_helpers.ex
# Copyright 2015 Six Colors AB
# Licensed under Apache 2.0 available at https://www.apache.org/licenses/LICENSE-2.0

defmodule BeaconWeb.Cache do
...
end
```

bcardarella:
Eric has given permission so we're good
Screenshot 2023-06-28 at 1 40 45 PM

PR #271, opened 2023-06-28
fix: remove requirement on :stylesheet_urls from Layout schema

### LINKED ISSUE #150: Make stylesheets optional
The minimum required to create a page should be a layout. But currently a stylesheet is also required, which shouldn't be.

See https://github.com/BeaconCMS/beacon/blob/bb2627ea484dbd2ef177e66ba59edd2adaae154b/test/beacon_web/live/page_live_test.exs#L154-L155 and https://github.com/BeaconCMS/beacon/blob/bb2627ea484dbd2ef177e66ba59edd2adaae154b/lib/beacon/layouts/layout.ex#L32-L33

polvalente:
closes #150 

It turns out that the stylesheets were already optional as the changeset was configured.

An additional validation through either `validate_change` or a combination of `get_field`/`get_change` and `add_error` must be done if an array field is to be considered invalid on an empty list, but this is not a concern in any of the other schemas which I found in the code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ beacon-*.tar
 /assets/node_modules
 # tailwind.install
 /assets/css/app.css
+/assets/tailwind.config.js
 
 # Dev temp files
 /dev/
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,9 @@
+version: "3.9" # optional since v1.27.0
+services:
+  db:
+    image: postgres:13.1
+    ports:
+      - "5432:5432"
+    environment:
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -35,6 +35,7 @@ beacon-*.tar
 /assets/node_modules
 # tailwind.install
 /assets/css/app.css
+/assets/tailwind.config.js
```
polvalente, line 38: This file was added during the setup it seems, and I'd accidentally commited it

docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+version: "3.9" # optional since v1.27.0
```
polvalente, line 1: Let me know if you'd rather not have docker-compose.yml in the project

polvalente, line 1:
```suggestion
version: "3.9"
```

leandrocp: That's fine, we need to support the 3 official adapters and having a docker-compose helps.

leandrocp:
Thanks!

PR #272, opened 2023-06-28
refactor: re-raise on DataSource.page_title/5

### LINKED ISSUE #157: Core -> Extension: page_title/2 callback silently swallows errors
If the user-provided data source raises an error in its implementation of `Beacon.DataSource.Behaviour.page_title/2`, there is no indication that an exception has occurred:

https://github.com/BeaconCMS/beacon/blob/7790eb72769a026c0bdfc3167aab394a9b73ce91/lib/beacon/data_source.ex#L46-L47

We want to change that to `reraise` the exception `Beacon.DataSourceError` providing a meaningful message similar to https://github.com/BeaconCMS/beacon/blob/7790eb72769a026c0bdfc3167aab394a9b73ce91/lib/beacon/lifecycle.ex#L163-L168

polvalente:
closes #157

### DIFF
--- a/lib/beacon/data_source.ex
+++ b/lib/beacon/data_source.ex
@@ -44,7 +44,16 @@ defmodule Beacon.DataSource do
       page_title
     end
   rescue
-    _error -> page_title
+    error ->
+      stacktrace = __STACKTRACE__
+
+      error_message = """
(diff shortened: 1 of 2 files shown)


leandrocp (changes requested):

leandrocp:
Thanks!

Issue #190, opened 2023-04-19
Admin -> General Improvements: Separate Beacon.Admin namespace into its own app `beacon_admin`


leandrocp:
For organization purposes and because these apps will have separated release management. Also this is a good opportunity to reorganize Beacon core modules into a well defined public API.

- [x] https://github.com/BeaconCMS/beacon/pull/282
- [x] https://github.com/BeaconCMS/beacon/pull/280
- [x] https://github.com/BeaconCMS/beacon/pull/278

Update: created https://github.com/BeaconCMS/beacon_live_admin

AZholtkevych:
Depends on #227

PR #280, opened 2023-07-06
Improvements to MediaLibrary

leandrocp:
No features have been changed.

- [x] Move `Beacon.Admin.MediaLibrary` to `Beacon.MediaLibrary` since admin namespace will be replaced by https://github.com/BeaconCMS/beacon_live_admin
- [x] Added some new functions in the `MediaLibrary` public API that takes `site` as argument which will be passed by `Beacon.LiveAdmin` which now is tenant-aware, so that means the media library will be scoped by site in beacon_live_admin but the existing admin in Beacon remains the same displaying all images from all sites.


xn (approved):

xn:
Excellent work!

leandrocp:
Thanks @xn!

leandrocp:
Related to https://github.com/BeaconCMS/beacon/issues/190

PR #282, opened 2023-07-10
Config.router -> Config.endpoint

leandrocp:
So we can generate the proper URL for assets


leandrocp:
https://github.com/BeaconCMS/beacon/issues/190

Issue #207, opened 2023-04-25
MANAGING ASSETS: Pages using assets


AZholtkevych:

AZholtkevych:
Able to show which pages are using a given asset, maybe even a grouping feature

PR #283, opened 2023-07-11
implement srcset support for the image component.

xn:
- adds source_id to Asset (indexed)
- adds usage_tag to Asset (indexed)
- adds has_many :assets to Asset
- adds has_one :thumbnail to Asset
- added image_set component


APB9785 (approved):
Looks good

dev.exs
```diff
@@ -216,7 +241,7 @@ seeds = fn ->
         <%= my_component("sample_component", val: 1) %>
 
         <div>
-          <BeaconWeb.Components.image name="dockyard_1.webp" width="200px" />
+          <BeaconWeb.Components.image_set asset={@beacon_live_data[:img1]} sources={["480w"]} width="200px" />
```
leandrocp, line 244: Hey @xn all looks good but I'm wondering if we can make an extra UX improvement here by replacing `asset` with `name` as:

```heex
<BeaconWeb.Components.image_set name="dockyard_1" sources={["480w"]} width="200px" />
```

Ideally components should avoid communication with external services, even Beacon.Repo, but on the other hand requiring a `DataSource` is counter productive for content creators that most likely don't know Elixir. Would be better to let users upload assets and maybe provide the code snippet so they can just copy & paste into the templates. And it's done. Wdyt?

leandrocp (commented):
Hey @xn good one as usual. I left just a comment regarding a possible UX improvement.

dev.exs
xn: This is a POC PR. I don't expect `image_set` to survive as is. There is a larger story about connecting the Media Library with the Page Editor and with Miguel's upcoming work, that will be all in flux. As I talked about in standup today, we can generate both markdown and heex snippets in the media library for use in the page editor and that will get us a good way into that story. The other side of it is connecting these assets to custom page attributes, like the illustration attr in dy.com.

Relying on the name is very brittle. We already have a ticket to be able to change the name of an asset and have no way to back propagate that to the existing usages. Not only that, it also gets complex when it interacts with the multiple backend feature. Also, if we rely solely on the name, then the only way we can figure out where a given asset is used is by a full text search and that won't be fun to write or debug.

In the end, I don't want to solve this problem in this PR, because it's a big deal that needs to interact with a bunch of moving targets and we need to think this one through quite thoroughly.

leandrocp: > be able to change the name of an asset and have no way to back propagate that to the existing usages

For reference, this problem is related to https://github.com/BeaconCMS/beacon/issues/207

PR #287, opened 2023-07-18
Validate page templates

### LINKED ISSUE #74: Core -> Route and Resource Loading: Handle runtime compilation errors

leandrocp:
Close #74

Screenshot 2023-07-18 at 5 30 19 PM

### DIFF
--- a/lib/beacon/content.ex
+++ b/lib/beacon/content.ex
@@ -33,17 +33,27 @@ defmodule Beacon.Content do
 
     case Beacon.Template.HEEx.compile(template, metadata) do
       {:cont, _ast} ->
-        changeset
+        {:ok, changeset}
 
       {:halt, %{description: description}} ->
-        Changeset.add_error(changeset, :template, description)
+        {:error, Changeset.add_error(changeset, :template, description)}
 
       {:halt, _} ->
-        Changeset.add_error(changeset, :template, "invalid template")
+        {:error, Changeset.add_error(changeset, :template, "invalid template")}
     end
   end
 
-  defp do_validate_page_template(changeset, _format, _template), do: changeset
+  defp do_validate_page_template(changeset, :markdown = _format, template) when is_binary(template) do
+    case Beacon.Template.Markdown.convert_to_html(template, %{}) do
+      {:cont, _template} ->
+        {:ok, changeset}
+
+      {:halt, _} ->
+        {:error, Changeset.add_error(changeset, :template, "invalid template")}
+    end
+  end
+
+  defp do_validate_page_template(changeset, _format, _template), do: {:ok, changeset}
 
   @doc """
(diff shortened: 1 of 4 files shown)


lib/beacon_web/live/admin/page_live/page_editor_live.ex
```diff
@@ -93,7 +93,7 @@ defmodule BeaconWeb.Admin.PageEditorLive do
         {:noreply, put_flash(socket, :info, "Page updated successfully")}
 
       {:error, {socket, _page}} ->
-        {:noreply, put_flash(socket, :info, "Failed to updated page")}
+        {:noreply, put_flash(socket, :error, "Failed to updated page")}
```
APB9785, line 96:
```suggestion
        {:noreply, put_flash(socket, :error, "Failed to update page")}
```

lib/beacon/content.ex
```diff
@@ -20,6 +20,31 @@ defmodule Beacon.Content do
...
+    case Beacon.Template.HEEx.compile(template, metadata) do
+      {:cont, _ast} ->
+        changeset
+
+      {:halt, %{description: description}} ->
+        Changeset.add_error(changeset, :template, description)
+
+      {:halt, _} ->
+        Changeset.add_error(changeset, :template, "invalid template")
+    end
```
APB9785, line 43: Perhaps this function should return a tagged tuple?  i.e. for error cases `{:error, changeset}` otherwise `{:ok, changeset}`

That way our later `with` blocks would not require `else` clause

lib/beacon/content.ex
```diff
@@ -367,17 +398,17 @@ defmodule Beacon.Content do
...
+
+      with %Changeset{valid?: true} <- validate_page_template(changeset),
+           {:ok, page} <- Repo.insert(changeset),
            {:ok, _event} <- create_page_event(page, "created"),
            %Page{} = page <- Lifecycle.Page.after_create_page(page) do
         {:ok, page}
+      else
+        %Changeset{} = changeset -> {:error, changeset}
+        error -> error
       end
```
APB9785, line 412: It would be nice if we did not need `else` here.  If we could do
```elixir
with {:ok, changeset} <- validate_page_template,
```
then the error case would already be tagged 🙂

leandrocp: Yep good catch. I'll make the change thanks

Issue #159, opened 2023-03-13
Core -> Hosting and Tenant: Compress CSS and JSS using Brotli


leandrocp:
We want to compress assets using [brotli](https://web.dev/reduce-network-payloads-using-text-compression/) which is [widely supported](https://caniuse.com/?search=brotli) and performs better than gzip.

Currently Beacon compress css and jss using gzip:

https://github.com/BeaconCMS/beacon/blob/7790eb72769a026c0bdfc3167aab394a9b73ce91/lib/beacon/runtime_css.ex#L27

https://github.com/BeaconCMS/beacon/blob/7790eb72769a026c0bdfc3167aab394a9b73ce91/lib/beacon/runtime_js.ex#L45

And serve those assets with proper content encoding:
 https://github.com/BeaconCMS/beacon/blob/7790eb72769a026c0bdfc3167aab394a9b73ce91/lib/beacon_web/controllers/assets_controller.ex#L23

Replace with brotli.

Ref https://github.com/BeaconCMS/beacon/pull/158

stefholbrook:
Looking into this 👀

PR #290, opened 2023-07-21
Brotli compression

stefholbrook:
This PR adds Brotli for css and js compression. 

This implementation uses  the [Brotli](https://hex.pm/packages/brotli) package. This seemed to fit our needs, had a straightforward implementation, and was the most widely used. 

Issue: [159](https://github.com/BeaconCMS/beacon/issues/159)


lib/beacon/runtime_css.ex
```diff
@@ -21,10 +21,17 @@ defmodule Beacon.RuntimeCSS do
...
-    true = :ets.insert(:beacon_assets, {{site, :css}, {hash, css, compressed}})
-    :ok
+
+    case :brotli.encode(css) do
+      {:ok, compressed} ->
+        hash = Base.encode16(:crypto.hash(:md5, css), case: :lower)
+        true = :ets.insert(:beacon_assets, {{site, :css}, {hash, css, compressed}})
+        :ok
+
+      error ->
```
stefholbrook, line 31: Not sure if we care about the error here, but included it just in case. Happy to remove it.

.github/workflows/elixir.yml
```diff
@@ -21,9 +21,11 @@ jobs:
         include:
           - elixir: "1.13.0"
             otp: "23"
+            rebar3_version: '3.15.2'
```
leandrocp, line 24: Do we still need these changes?

lib/beacon/runtime_css.ex
leandrocp: That's fine. Raising here means that a site wouldn't load but on the other hand starting a site without styles and JS isn't a good option either.
The alternative is fall backing to gzip or uncompressed but it's not worth it unless we start having issues.

priv/static/beacon_admin.css
leandrocp, line 1: Can you please revert the changes on these static files?

leandrocp (changes requested):
Nice one thanks!

xn (approved):

leandrocp (approved):
Thanks! 🎉

PR #294, opened 2023-07-26
fix file input accept attribute validation

xn:
```
<input type="file" accept=".pdf, image/jpeg, video/*" >
```
is valid.

previous to this PR we only accepted `accept` in the format:
```
<input type="file" accept="application/pdf, image/jpeg, video/mpeg" >
```

Had to change the logic of building default configs to accommodate this.
Also, now the default processor can handle any arbitrary file type.

### DIFF
--- a/lib/beacon/config.ex
+++ b/lib/beacon/config.ex
@@ -57,7 +57,7 @@ defmodule Beacon.Config do
   @typedoc """
   Register specific media types allowed for upload. Catchalls are not allowed.
   """
-  @type allowed_media_types :: [media_type :: String.t()]
+  @type allowed_media_accept_types :: [media_type :: String.t()]
 
   @typedoc """
   Register backends and validations for media types. Catchalls are allowed.
@@ -136,7 +136,7 @@ defmodule Beacon.Config do
           safe_code_check: safe_code_check(),
           template_formats: template_formats(),
           assets: assets(),
-          allowed_media_types: allowed_media_types(),
+          allowed_media_accept_types: allowed_media_accept_types(),
           lifecycle: lifecycle(),
           extra_page_fields: extra_page_fields(),
           default_meta_tags: default_meta_tags()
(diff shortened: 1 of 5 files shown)


lib/beacon/config.ex
```diff
@@ -422,36 +422,55 @@ defmodule Beacon.Config do
...
+        media_type
+    end
   end
 
-  defp process_assets_config(allowed_media_types, assigned_assets) do
+  # .some_ext (not checking for valid extension)
+  defp validate_accept_extension(<<46, _rest::binary>> = media_type), do: media_type
+
+  # catchall
+  defp validate_accept_extension(media_type), do: raise("`#{media_type}` does not appear to be a media type, extensions must begin with a `.`")
```
leandrocp, line 463: You most likely want to raise LoaderError here

```suggestion
  defp validate_accept_extension(media_type), do: raise Beacon.LoaderError, "`#{media_type}` does not appear to be a media type, extensions must begin with a `.`"
```

lib/beacon/config.ex
```diff
@@ -466,6 +485,32 @@ defmodule Beacon.Config do
...
+      configs ++ [{media_type, [{:backends, [Beacon.MediaLibrary.Backend.Repo]}]}]
+    end
+  end
+
+  defp ensure_backend_for_extension(configs, <<46, extension::binary>>) do
+    if MIME.has_type?(extension) do
+      media_type = MIME.type(extension)
+      ensure_backend(configs, media_type)
+    else
+      raise """
```
leandrocp, line 505:
```suggestion
      raise Beacon.LoaderError, """
```

lib/beacon/config.ex
```diff
@@ -466,6 +485,32 @@ defmodule Beacon.Config do
...
+      No known media type for: #{extension}
+      """
+    end
+  end
+
+  defp ensure_backend_for_extension(_configs, extension_without_leading_dot),
+    do: raise("`#{extension_without_leading_dot}` does not appear to be a media type, extensions must begin with a `.`")
```
leandrocp, line 512:
```suggestion
    do: raise Beacon.LoaderError, "`#{extension_without_leading_dot}` does not appear to be a media type, extensions must begin with a `.`"
```

leandrocp (changes requested):
Nice!

lib/beacon/config.ex
xn: yes!

leandrocp (approved):

PR #296, opened 2023-07-26
Add page variants

### LINKED ISSUE #288: Admin -> General Improvements: a/b/c/d/e/f/etc... testing spec
Beacon should be able to support multiple variants of a given URL and be able to set weights on the render frequency.

The variants should be infinite. When visiting the URL the variants are not present in the URL however we could selectively render a given variant with a query param:

`https://dockyard.com/campaign-page?variant=a`

something like that

for analytics and conversion testing we can embed the variant name.

APB9785:
Resolves #288 

This PR adds a new `PageVariant` table+schema, allowing users to define an alternate template for any given page, along with a "weight" which is the percentage of page renders which should use that template.


lib/beacon_web/live/page_live.ex
```diff
@@ -24,7 +24,11 @@ defmodule BeaconWeb.PageLive do
   end
 
   def render(assigns) do
-    {{site, path}, {_page_id, _layout_id, format, template, _page_module, _component_module}} = lookup_route!(assigns.__site__, assigns.__live_path__)
+    {{site, path}, {_page_id, _layout_id, format, templates, _page_module, _component_module}} =
+      lookup_route!(assigns.__site__, assigns.__live_path__)
+
+    template = choose_template(templates)
```
leandrocp, line 30: How do you see the UI for choosing the weights? If users choose A/40 and B/60 then will it display 0 for primary? Similarly if they choose A/10 and B/20 would it display 70 for primary?

lib/beacon_web/live/page_live.ex
```diff
@@ -37,6 +41,17 @@ defmodule BeaconWeb.PageLive do
       """
   end
 
+  defp choose_template(templates) do
```
leandrocp, line 44: You can return early here:

```suggestion
  defp choose_template(%{variants: [], primary: primary}), do: primary
  
  defp choose_template(templates) do
```

lib/beacon_web/live/page_live.ex
APB9785: Yes I think on a "Variants" page in `/admin` it would list each variant with an input for the weight.  We can include validation where `if sum(weights) > 100` then an error is displayed and changes cannot be submitted.  Then at the bottom (or top) of the variants page, it can show `Primary template weight` with live updates based on `100 - sum(weights)`.  Basically the same as in your example 🙂

leandrocp: Sounds good 👍🏻

lib/beacon_web/live/page_live.ex
APB9785: Yes, good idea.  Here we should take every opportunity to maximize performance since it will happen on every page render.

leandrocp (changes requested):
Nice! Found no issues, just a few suggestions.

lib/beacon/content.ex
```diff
@@ -1419,4 +1432,93 @@ defmodule Beacon.Content do
...
+  @doc """
+  Updates a page variant and returns the page with updated variants association.
+  """
+  @doc type: :page_variants
+  @spec update_variant_for_page(Page.t(), PageVariant.t(), map()) :: {:ok, Page.t()} | {:error, Changeset.t()}
+  def update_variant_for_page(page, variant, attrs) do
+    changeset = PageVariant.changeset(variant, attrs)
+
+    Repo.transact(fn ->
+      with {:ok, _} <- validate_variant_template(changeset, page),
```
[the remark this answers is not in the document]
APB9785: This was intentional but I agree it could be handled better.  Perhaps `{:ok, ^changeset}` would make it clear that the changeset is being returned unchanged in the success case

leandrocp (approved):
🎉

PR #299, opened 2023-07-28
create Beacon.MediaTypes.normalize/1

xn:
Browsers' media types are often out of date, returning and empty string or a deprecated assignation which the elixir MIME library no longer supports.

created Beacon.MediaTypes.normalize/1 which handles these cases

### DIFF
--- a/lib/beacon/media_types.ex
+++ b/lib/beacon/media_types.ex
@@ -1,0 +1,10 @@
+defmodule Beacon.MediaTypes do
+  # Browsers' media types are often out of date,
+  # returning and empty string or a deprecated assignation
+  # which the elixir MIME library no longer supports.
+  def normalize(""), do: nil
+
+  def normalize("application/font-woff"), do: "font/woff"
+
+  def normalize(media_type), do: media_type
+end
--- a/lib/beacon/media_types.ex
+++ b/lib/beacon/media_types.ex
@@ -14,8 +14,6 @@ defmodule Beacon.MediaTypes do
   which the Elixir MIME library no longer supports.
   """
   def normalize(""), do: nil
-
   def normalize("application/font-woff"), do: "font/woff"
-
   def normalize(media_type), do: media_type
 end
--- a/lib/beacon/media_types.ex
+++ b/lib/beacon/media_types.ex
@@ -9,9 +9,9 @@ defmodule Beacon.MediaTypes do
   """
 
   @doc """
-    Browsers' media types are often out of date,
-    returning and empty string or a deprecated media type
-    which the elixir MIME library no longer supports.
+  Browsers' media types are often out of date,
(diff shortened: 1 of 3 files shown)


lib/beacon/media_types.ex
```diff
@@ -0,0 +1,10 @@
+defmodule Beacon.MediaTypes do
+  # Browsers' media types are often out of date,
```
leandrocp, line 2: You can turn this comment into an actual `@doc`

leandrocp (approved):
Just one small suggestion

lib/beacon/media_types.ex
```diff
@@ -0,0 +1,21 @@
+defmodule Beacon.MediaTypes do
+  @moduledoc """
+    #{__MODULE__} serves as a context to encapsulate business logic
+    around media types that is not handled by the MIME library.
+
+    Primary concerns:
+    - Convenience functions
+    - Functions to handle edge cases between the browser and elxir
+  """
```
APB9785, line 9:
```suggestion
  @moduledoc """
  #{__MODULE__} serves as a context to encapsulate business logic
  around media types that is not handled by the MIME library.

  Primary concerns:
    - Convenience functions
    - Functions to handle edge cases between the browser and Elixir
  """
```

lib/beacon/media_types.ex
```diff
@@ -0,0 +1,21 @@
...
+
+  @doc """
+    Browsers' media types are often out of date,
+    returning and empty string or a deprecated media type
+    which the elixir MIME library no longer supports.
+  """
```
APB9785, line 15:
```suggestion
  @doc """
  Browsers' media types are often out of date,
  returning an empty string or a deprecated media type
  which the Elixir MIME library no longer supports.
  """
```

lib/beacon/media_types.ex
```diff
@@ -0,0 +1,21 @@
...
+  def normalize(""), do: nil
+
+  def normalize("application/font-woff"), do: "font/woff"
+
+  def normalize(media_type), do: media_type
```
APB9785, line 20:
```suggestion
  def normalize(""), do: nil
  def normalize("application/font-woff"), do: "font/woff"
  def normalize(media_type), do: media_type
```

lib/beacon_web/live/admin/media_library_live/upload_form_component.ex
```diff
@@ -83,7 +83,8 @@ defmodule BeaconWeb.Admin.MediaLibraryLive.UploadFormComponent do
           site -> site
         end
 
-      uploaded_assets =
+      media_type =
+        uploaded_assets =
```
APB9785, line 87:
```suggestion
      uploaded_assets =
```

PR #303, opened 2023-07-31
Convert between snake_case and camelCase automatically.

### LINKED ISSUE #300: UI Builder API: Ensure elixir works with snake_case and sends camelCase to the consumers
We should be able to use a plug to automatically convert between them so each works uses their standard. 

This could do the trick: https://github.com/malomohq/accent

cibernox:
Closes #300

Internally elixir uses snake_case consistently, but most JS APIs send and receive camelCase. A new plug makes the translation when sending or receiving transparently.

### DIFF
--- a/dev.exs
+++ b/dev.exs
@@ -58,6 +58,8 @@ defmodule SamplePhoenixWeb.Router do
 
   pipeline :api do
     plug :accepts, ["json"]
+    plug Accent.Plug.Request, default_case: Accent.Case.Snake
+    plug Accent.Plug.Response, default_case: Accent.Case.Camel, json_codec: Jason
   end
 
   scope "/" do
--- a/dev.exs
+++ b/dev.exs
@@ -59,8 +59,6 @@ defmodule SamplePhoenixWeb.Router do
   pipeline :api do
     plug :accepts, ["json"]
     plug BeaconWeb.API.Plug
-    # plug Accent.Plug.Request, default_case: Accent.Case.Snake
-    # plug Accent.Plug.Response, default_case: Accent.Case.Camel, json_codec: Jason
   end
 
   scope "/" do
--- a/dev.exs
+++ b/dev.exs
@@ -58,8 +58,9 @@ defmodule SamplePhoenixWeb.Router do
 
   pipeline :api do
     plug :accepts, ["json"]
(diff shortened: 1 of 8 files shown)


lib/beacon/blueprint_converter.ex
```diff
@@ -55,6 +55,9 @@ defmodule Beacon.BlueprintConverter do
   defp render_node(node) when is_binary(node), do: node
   # Special case to just output content as HTML
   defp render_node(%{"tag" => "raw", "content" => content}), do: content
+  defp render_node(%{"tag" => "img", "attributes" => attributes}) do
```
cibernox, line 58: This is unrelated but I found a bug with pages that contained `` tags because those should be self-closed.

dev.exs
```diff
@@ -58,6 +58,8 @@ defmodule SamplePhoenixWeb.Router do
 
   pipeline :api do
     plug :accepts, ["json"]
+    plug Accent.Plug.Request, default_case: Accent.Case.Snake
+    plug Accent.Plug.Response, default_case: Accent.Case.Camel, json_codec: Jason
```
leandrocp, line 62: If we go down this route we'll need to:
- create a `BeaconWeb.API.Plug` as https://hexdocs.pm/plug/Plug.Builder.html
- update https://github.com/BeaconCMS/beacon/blob/main/guides/introduction/api.md to instruct users to add `plug BeaconWeb.API.Plug` in the `pipeline :api` block in theirs apps' router.
- and then replace these 2 lines with `plug BeaconWeb.API.Plug` as well

cibernox: Ready. It's way neater!

leandrocp: Yeah and the main reason for that change is to avoid exposing 3rd-party code that we have no control over.

dev.exs
```diff
@@ -58,6 +58,9 @@ defmodule SamplePhoenixWeb.Router do
 
   pipeline :api do
     plug :accepts, ["json"]
+    plug BeaconWeb.API.Plug
+    # plug Accent.Plug.Request, default_case: Accent.Case.Snake
+    # plug Accent.Plug.Response, default_case: Accent.Case.Camel, json_codec: Jason
```
leandrocp, line 63: You can remove these lines now :)

```suggestion
```

leandrocp (approved):

PR #312, opened 2023-08-07
Move layout and page template validation to context

leandrocp:
The same pattern used on page variants, which is more effective since it only validates when necessary and not on every change.


APB9785 (approved):
Looks good

PR #314, opened 2023-08-07
Add delete variant function

APB9785:


lib/beacon/content.ex
```diff
@@ -1572,4 +1572,17 @@ defmodule Beacon.Content do
 
   # TODO: expose template validation to custom template formats defined by users
   defp do_validate_template(changeset, _field, _format, _template, _metadata), do: {:ok, changeset}
+
+  @doc """
+  Deletes a page variant and returns the page with updated variants association.
+  """
+  @doc type: :page_variants
+  @spec delete_variant_from_page(Page.t(), PageVariant.t()) :: {:ok, Page.t()} | {:error, Changeset.t()}
+  def delete_variant_from_page(page, variant) do
```
leandrocp, line 1581: How are we gonna handle published variants?

1. If we update the ETS table right after deleting a variant then it would stop rendering that variant, which may be the expected behavior, but it differs a bit from the existing publish workflow that requires publishing a page whenever there's a change in the page.
2 - Postpone updating the ETS table only when the page gets published eventually, which aligns with the existing workflow but may force the user to publish a version that's not ready yet just to get rid of a variant. In the other hand deleting a variant may be done mostly when users have finished A/B testing and really wanna update the main template and publish it.

Since that's related to managing variants individually, which is a broader discussion and has plenty of trade-offs, I'm leaning towards the second option _but_ adding a friendly warning to users know the page has to be published to have any effect.

Wdyt?

PR #315, opened 2023-08-07
Validate total variant weights

APB9785:
Regardless of template validation result, also check the total weights and if it's > 100, add an error to the changeset.


lib/beacon/content.ex
```diff
@@ -1549,11 +1549,48 @@ defmodule Beacon.Content do
...
+      {:ok, changeset} ->
+        do_validate_weights(changeset, page)
+
+      {:error, changeset} ->
+        {_, updated_changeset} = do_validate_weights(changeset, page)
+        {:error, updated_changeset}
+    end
+  end
+
+  defp do_validate_weights(changeset, page) do
```
leandrocp, line 1567: Can we have a test for this logic?

leandrocp (approved):

PR #323, opened 2023-08-09
Layout external resource links

### LINKED ISSUE #322: Augment external resource links options
Currently Beacon supports adding [`stylesheet_urls` to layouts](https://github.com/BeaconCMS/beacon/blob/6389ce57f477eba9bb14d980b173ad845794578d/lib/beacon/content/layout.ex#L24) which are [injected into the `<head>`](https://github.com/BeaconCMS/beacon/blob/6389ce57f477eba9bb14d980b173ad845794578d/lib/beacon_web/components/layouts/runtime.html.heex#L10) but we need to:

1 - Support more options besides `rel="stylesheet"`. For example to load fonts as:

```html
   <link
    rel="preload"
    href="https://mycdn.com/manrope-v13-latin-latin-ext-regular.woff2"
    as="font"
    type="font/woff2"
    crossorigin="anonymous" />
```

2 - Allow editing that field in LiveAdmin, see https://github.com/BeaconCMS/beacon_live_admin/issues/41

---

The implementation will be very similar to meta tags where we have a list of maps and each attribute of `<link>` is defined as key/value of that map, for example the link above would look like:

```elixir
%Layout{
  resource_links: [
    %{
      rel: "preload",
      href: "https://mycdn.com/manrope-v13-latin-latin-ext-regular.woff2",
      as: "font",
      type: "font/woff2",
      crossorigin: "anonymous"
    }
  ]
}
```

### Tasks

- [ ] Rename [`stylesheet_urls`](https://github.com/BeaconCMS/beacon/blob/6389ce57f477eba9bb14d980b173ad845794578d/lib/beacon/content/layout.ex#L34) to `resource_links`
- [ ]  Change that field data type from `[String.t]` to `[map()]`
- [ ] Create a migration to migrate existing data to the new data structure
- [ ] Rename [linked_stylesheets/1](https://github.com/BeaconCMS/beacon/blob/6389ce57f477eba9bb14d980b173ad845794578d/lib/beacon_web/components/layouts/runtime.html.heex#L10) and change it to render each `<link>` with attrs
- [ ] Bump layout version similar to https://github.com/BeaconCMS/beacon/commit/6389ce57f477eba9bb14d980b173ad845794578d

### Refs
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link

leandrocp:
Close #322

### DIFF
--- a/priv/repo/migrations/20230809190953_add_resource_links_to_layouts.exs
+++ b/priv/repo/migrations/20230809190953_add_resource_links_to_layouts.exs
@@ -7,8 +7,9 @@ defmodule Beacon.Repo.Migrations.AddResourceLinksToLayouts do
     end
 
     for layout_id <- layout_ids() do
-      links = stylesheet_links_for_layout(layout_id)
-      update_layout_links(layout_id, links)
+      layout_id
+      |> stylesheet_links_for_layout()
+      |> update_layout_links(layout_id)
     end
 
      alter table(:beacon_layouts) do
@@ -57,7 +58,7 @@ defmodule Beacon.Repo.Migrations.AddResourceLinksToLayouts do
     end
   end
 
-  defp update_layout_links(layout_id, links) do
+  defp update_layout_links(links, layout_id) do
     execute(fn ->
       repo().query!(
         """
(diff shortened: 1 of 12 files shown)


priv/repo/migrations/20230809190953_add_resource_links_to_layouts.exs
```diff
@@ -0,0 +1,80 @@
...
+    query = """
+    SELECT DISTINCT id FROM beacon_layouts
+    """
+
+    case repo().query(query, [], log: :info) do
+      {:ok, result} ->
+        for row <- result.rows do
+          %{id: id} = repo().load(%{id: :binary_id}, {result.columns, row})
+          {:ok, id} = Ecto.UUID.dump(id)
+          id
```
APB9785, line 30:
```suggestion
          Ecto.UUID.dump!(id)
```

APB9785 (approved):
Nice

PR #325, opened 2023-08-10
Reload component on update

### LINKED ISSUE #324: Core -> Route and Resource Loading: reload component on update
When a component is updated it should be reloaded with the following code to apply the changes on the site:

```elixir
    Beacon.Loader.ComponentModuleLoader.load_components(
      current_site,
      Beacon.Content.list_components(current_site)
    )
```

Currently there's no way to reload single components, it has to reload all of them at once.

leandrocp:
Close #324

Related to https://github.com/BeaconCMS/beacon_live_admin/issues/16

### DIFF
--- a/lib/beacon/content.ex
+++ b/lib/beacon/content.ex
@@ -1459,9 +1459,9 @@ defmodule Beacon.Content do
     |> Repo.update()
     |> tap(&maybe_reload_component/1)
   end
-  
-  def maybe_update_component({:ok, component}), do: PubSub.component_updated(component)
-  def maybe_update_component({:error, _component}), do: :noop
+
+  def maybe_reload_component({:ok, component}), do: PubSub.component_updated(component)
+  def maybe_reload_component({:error, _component}), do: :noop
 
   defp validate_component_body(changeset) do
     site = Changeset.get_field(changeset, :site)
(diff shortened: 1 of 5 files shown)


lib/beacon/content.ex
```diff
@@ -1457,6 +1457,14 @@ defmodule Beacon.Content do
...
     |> Repo.update()
+    |> case do
+      {:ok, component} ->
+        :ok = PubSub.component_updated(component)
+        {:ok, component}
+
+      error ->
+        error
+    end
   end
```
APB9785, line 1468:
```suggestion
    |> tap(&maybe_reload_component/1)
  end
  
  def maybe_update_component({:ok, component}), do: PubSub.component_updated(component)
  def maybe_update_component({:error, _component}), do: :noop
```
This might be controversial, but WDYT about `tap/2` to make it clear that the result of `Repo.update/2` will not be changed?  I think I slightly prefer it, but either way is fine, so feel free to disregard this if you don't agree 🙂

APB9785 (approved):

lib/beacon/content.ex
leandrocp: That's not a pattern I've seen used in other projects too much but I'm willing to give it a try

PR #326, opened 2023-08-10
Core -> Page Builder: Implement rendering of templates in the page's API [PART 1]

cibernox:
### What's included in this PR
This allows to send and receive not the page's raw HEEx template, but the the AST representing that template in a json-friendly format.
Likewise, when saving a page, it receives the that json-ast and transforms it back to an HEEx template.

The UI builder has been updated to leverage that AST for a much simpler interaction model.

### What's left for part 2?
The AST contains certain nodes, like :eex and :eex_blocks that the client will have no idea how to render. For those, we will likely need to expand the AST to include the rendered version of that component or expression (rendered with sample data I presume) that the UI builder will display verbatim.


test/beacon/template/heex_transformer_test.exs
```diff
@@ -0,0 +1,35 @@
...
+          }
+        ],
+        "attrs" => %{}
+      }
+    ]
+
+    # heex_code = Phoenix.LiveView.HTMLFormatter.format(HeexTransformer.transform(json_ast), heex_line_length: 100)
+    heex_code = HeexTransformer.transform(json_ast)
+
+    assert heex_code ==
```
leandrocp, line 32: that's awesome

lib/beacon/content/page.ex
```diff
@@ -99,8 +100,16 @@ defmodule Beacon.Content.Page do
...
+  def update_changeset(page, attrs \\ %{}) do
     {extra_attrs, attrs} = Map.pop(attrs, "extra")
+    {ast, attrs} = Map.pop(attrs, "ast")
+
+    attrs =
+      if is_nil(ast) do
+        attrs
+      else
+        Map.put(attrs, :template, HeexTransformer.transform(ast))
+      end
```
leandrocp, line 112: Instead of manipulating `attrs` in the changeset, it's better to move this block of code to https://github.com/BeaconCMS/beacon/blob/47490d983b99f92b893b0a9cfa55aac1b3aa3c14/lib/beacon/content.ex#L485 so the schema doesn't need to know about `ast`, it only cares about `template`.

lib/beacon/template/heex_transformer.ex
```diff
@@ -0,0 +1,47 @@
+defmodule Beacon.Template.HEEx.HeexTransformer do
+  def transform(ast) do
+    :erlang.iolist_to_binary(_transform(ast))
```
leandrocp, line 3:
```suggestion
    ast
    |> _transform()
    |> :erlang.iolist_to_binary()
```

lib/beacon/template/heex_transformer.ex
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  defp _transform([str]) when is_binary(str), do: str
+
+  defp _transform(ast) do
+    ast
+    |> Enum.map(&transform_node/1)
```
leandrocp, line 10:
```suggestion
    Enum.map(ast, &transform_node/1)
```

lib/beacon/template/heex_transformer.ex
```diff
@@ -0,0 +1,47 @@
+defmodule Beacon.Template.HEEx.HeexTransformer do
+  def transform(ast) do
```
leandrocp, line 2: I think it should format the resulting heex code. Users will edit the template in the admin code editor and page builder at the same time, while the code in the code editor will most likely be formatted (or close to it), editing a page in the page builder will result in unformatted code causing a bad UX.

It's not a blocking issue for this PR but I think would be useful to call https://github.com/phoenixframework/phoenix_live_view/blob/fb111738d56745f37338867b9faea86eb9baa6e1/lib/phoenix_live_view/html_formatter.ex#L233

leandrocp (changes requested):
It's awesome to see the 2-way integration after all the discussions we had about this design. Nice work.

lib/beacon/template/heex_transformer.ex
cibernox: Done

lib/beacon/content/page.ex
cibernox: Done

leandrocp (approved):
🎉

lib/beacon/blueprint_converter.ex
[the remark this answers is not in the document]
cibernox: Done

PR #331, opened 2023-08-17
Error Pages

### LINKED ISSUE #137: Core -> Route and Resource Loading: Support error pages
Phoenix Endpoint [renders templates](https://hexdocs.pm/phoenix/custom_error_pages.html) for errors given the [`Endpoint :render_errors`](https://hexdocs.pm/phoenix/Phoenix.Endpoint.html#module-runtime-configuration) is properly set, and Beacon integrates with this mechanism by raising [custom exceptions](https://github.com/BeaconCMS/beacon/blob/main/lib/beacon_web/exceptions.ex) with a proper `:plug_status`.

Currently the only supported way to render these error page is by defining them in the host application, but we want to allow users to write the templates for 4xx, 5xx errors at runtime.

### Beacon
- Add table `beacon_error_pages`
- Add schema `ErrorPage` with at least `:template` and `:status` fields. Status must be one of the 4xx and 5xx as listed at https://github.com/elixir-plug/plug/blob/main/lib/plug/conn/status.ex#L8
- Functions in `Beacon.Content` to manage these error pages. A good example is https://github.com/BeaconCMS/beacon/pull/328
- Change the [install script](https://github.com/BeaconCMS/beacon/blob/main/lib/mix/tasks/install.ex) to modify the [Endpoint :render_errors config](https://hexdocs.pm/phoenix/Phoenix.Endpoint.html#module-runtime-configuration) value to `html: BeaconWeb.ErrorHTML`
- Compile a "error module" per site containing all these error templates in `Beacon.Loader`. Similar to how Layouts are loaded.
- In `BeaconWeb.ErrorHTML` is must pattern match some private assign to find the source site and call the correct render function from the error module for that site. This module acts as a proxy. That's the part we're unsure and should do some research.

### Beacon LiveAdmin
- Add a new menu item "Error Pages" where users can create a template for each available status. A good example is https://github.com/BeaconCMS/beacon_live_admin/pull/50

APB9785:
## Goal

Beacon should allow creation of error pages at runtime, which are used for rendering HTTP error responses
Resolves #137 

## Implementation

* Beacon now seeds a Default layout, as well as 404 and 500 error pages
* Error pages are unique by site + status code
* Only valid error status codes are allowed
* `mix beacon.install` now sets the endpoint's :render_errors config


lib/beacon/content.ex
```diff
@@ -1667,6 +1680,108 @@ defmodule Beacon.Content do
...
+  def change_error_page(%ErrorPage{} = error_page, attrs \\ %{}) do
+    ErrorPage.changeset(error_page, attrs)
+  end
+
+  @doc """
+  Returns the error page for a given site and status code, or `nil` if no matching error page exists.
+  """
+  @doc type: :error_pages
+  @spec get_error_page(Site.t(), integer()) :: ErrorPage.t() | nil
+  def get_error_page(site, status) do
```
leandrocp, line 1705: Wdyt about being a bit more specific here?

```suggestion
  @type status_code :: 400..599

  @spec get_error_page(Site.t(), status_code()) :: ErrorPage.t() | nil
  def get_error_page(site, status) when status >= 400 and status <= 599 do
```

lib/beacon/content.ex
```diff
@@ -1667,6 +1680,108 @@ defmodule Beacon.Content do
...
+        where: e.site == ^site,
+        order_by: e.status
+    )
+  end
+
+  @doc """
+  Creates a new error page.
+  """
+  @doc type: :error_pages
+  @spec create_error_page(%{site: Site.t(), status: integer(), template: binary(), layout_id: Ecto.UUID.t()}) ::
```
leandrocp, line 1730: can use that type here too

```suggestion
  @spec create_error_page(%{site: Site.t(), status: status_code(), template: binary(), layout_id: Ecto.UUID.t()}) ::
```

lib/beacon/content.ex
```diff
@@ -1667,6 +1680,108 @@ defmodule Beacon.Content do
...
+    %ErrorPage{}
+    |> ErrorPage.changeset(attrs)
+    |> Repo.insert()
+  end
+
+  @doc """
+  Creates a new error page, raising if the operation fails.
+  """
+  @doc type: :error_pages
+  @spec create_error_page!(%{site: Site.t(), status: integer(), template: binary(), layout_id: Ecto.UUID.t()}) ::
```
leandrocp, line 1742:
```suggestion
  @spec create_error_page!(%{site: Site.t(), status: status_code(), template: binary(), layout_id: Ecto.UUID.t()}) ::
```

lib/beacon/content.ex
```diff
@@ -1667,6 +1680,108 @@ defmodule Beacon.Content do
...
+  @doc """
+  Returns attr data to load the default error_pages into new sites.
+  """
+  @spec default_error_pages() :: [map()]
+  @doc type: :error_pages
+  def default_error_pages do
+    for status <- [404, 500] do
+      %{
+        status: status,
+        template: Plug.Conn.Status.reason_phrase(status)
```
leandrocp, line 1760: 👍🏻

lib/mix/tasks/install.ex
```diff
@@ -91,6 +92,37 @@ defmodule Mix.Tasks.Beacon.Install do
     end
   end
 
+  @doc false
+  def inject_endpoint_render_errors_config(config_file_path) do
```
leandrocp, line 96: 🆒

test/support/fixtures.ex
```diff
@@ -185,4 +186,27 @@ defmodule Beacon.Fixtures do
...
+      site: attrs[:site] || "my_site",
+      status: attrs[:status] || Enum.random(111..999),
+      template: attrs[:template] || "Uh-oh!",
+      layout_id: attrs[:layout_id] || layout_fixture().id
+    }
+
+    %ErrorPage{}
+    |> ErrorPage.changeset(full_attrs)
+    |> Repo.insert!()
+  end
```
leandrocp, line 211: You could rewrite it as:

```suggestion
  def error_page_fixture(attrs \\ %{})
  end

  def error_page_fixture(%{layout: %Content.Layout{}} = attrs) do
    {layout, attrs} = Map.pop!(attrs, :layout)

    attrs
    |> Map.put(:layout_id, layout.id)
    |> error_page_fixture()
  end

  def error_page_fixture(attrs) do
    full_attrs = %{
      site: attrs[:site] || "my_site",
      status: attrs[:status] || Enum.random(111..999),
      template: attrs[:template] || "Uh-oh!",
      layout_id: attrs[:layout_id] || layout_fixture().id
    }

    %ErrorPage{}
    |> ErrorPage.changeset(full_attrs)
    |> Repo.insert!()
  end
```

leandrocp, line 211: you could rewrite as:

```suggestion
  def error_page_fixture(attrs \\ %{}) do
    layout_id = get_lazy(attrs, :layout_id, fn -> layout_fixture().id end)

    attrs
    |> Enum.into(%{
      site: "my_site",
      status: Enum.random(400..599),
      template: "Uh-oh!",
      layout_id: layout_id
    })
    |> Content.create_error_page!()
  end
```

this way you can call the fixture as `error_page_fixture(status: 404)` or with a map if you wish while keeping default values with less lines of code :)

lib/beacon_web/controllers/error_html.ex
```diff
@@ -1,21 +1,21 @@
...
+      |> Beacon.Content.get_error_page(status)
+      |> Beacon.Repo.preload(:layout)
 
-  # The default is to render a plain text page based on
-  # the template name. For example, "404.html" becomes
-  # "Not Found".
-  def render(template, _assigns) do
-    Phoenix.Controller.status_message_from_template(template)
+    EEx.eval_string(layout_template, assigns: [inner_content: page_template])
   end
```
leandrocp, line 20: I think we need to be more strict to avoid errors here, something like:

```elixir
  def render(
        <<status_code::binary-size(3), _rest::binary>>,
        %{
          conn: %{private: %{phoenix_live_view: {_, _, %{extra: %{session: %{"beacon_site" => site}}}}}}
        }
      ) do
    status_code
    |> String.to_integer()
    |> render_error_page(site)
  rescue
    _ -> render_default_error_page()
  end

  defp render_error_page(status, site) when status >= 400 and status <= 599 and is_atom(site) do
    error_page =
      site
      |> Beacon.Content.get_error_page(status)
      |> Beacon.Repo.preload(:layout)

    case error_page do
      %{layout: %{template: layout_template}, template: page_template} ->
        EEx.eval_string(layout_template, assigns: [inner_content: page_template])

      nil ->
        "render some default error page for the given status code"
    end
  end

  defp render_default_error_page do
    "render a default error page here"
  end
```

leandrocp:
I've submitted the first review incomplete by mistake so I had to submit others to finalize the review. But I'm done now :)
It's looking good, just left a few suggestions.

test/support/fixtures.ex
APB9785: The purpose of the extra clause here is to handle `error_page_fixture(%{layout: layout})`.  Almost all of the tests will have the full layout available, so we probably want to allow the fixture to be called that way.

leandrocp: Passing `layout` sounds good, in that case it can be `layout_id = get_lazy(attrs, :layout, fn -> layout_fixture() end).id` that will have the same result

lib/beacon_web/controllers/error_html.ex
APB9785: This would all be replaced with the module recompilation approach we discussed last week, right?

lib/beacon/content.ex
APB9785: Yeah in my first approach I was even more specific, with a list of only valid error codes (could later be used for form select).  I'll bring that back

lib/beacon_web/controllers/error_html.ex
leandrocp: Yes part of it would be replaced. If you intended to work on it then it's fine to keep the existing code since it will be replaced soon.

lib/beacon/content.ex
leandrocp: > Yeah in my first approach I was even more specific, with a list of only valid error codes (could later be used for form select). I'll bring that back

Yep that sounds good

PR #335, opened 2023-08-21
Reload layout

leandrocp:
Reloading a layout will force all pages using that layout to be recompiled without a template which will be reloaded on demand as requests try to render such pages, since we can't just recompile all page modules at once. It won't change the page the user is visiting, only on the next request or after refreshing the page.


APB9785 (approved):
Nice

PR #336, opened 2023-08-22
Fix publish process

leandrocp:
- Move the pubsub broadcast outside of the repo transaction since `Beacon.Loader` query the database to find the latest published page/layout.
- Revert `PageModuleLoader` as a simple module and let only `Loader` be a genserver so we have a single queue for loading resources since they depend on each other.
- Revert page template unloading, introduced at https://github.com/BeaconCMS/beacon/pull/335, since we require users to refresh the page to see the updated markup and the pubsub fix is all we need.

### DIFF
--- a/lib/beacon/content.ex
+++ b/lib/beacon/content.ex
@@ -133,18 +133,24 @@ defmodule Beacon.Content do
   Event + snapshot
   """
   @doc type: :layouts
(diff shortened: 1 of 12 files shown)


APB9785 (approved):
Looks good 🥳

PR #354, opened 2023-10-05
Fix bug in seeds file listing on installation.md

mattvanhorn:
The URL params will be passed with string keys, but the  line with `@beacon_path_params.blog_slug` is looking for an atom as the key, and so will fail with a KeyError, like so:
```
(KeyError) key :blog_slug not found in: %{"blog_slug" => "my_first_post"}
```
This took me a little while to figure out, so hopefully this will save someone some time, and come up when they google the error message.


leandrocp:
Thanks!

PR #363, opened 2023-10-20
Live Data Source

APB9785:
### Summary

This PR adds everything needed in the Beacon backend to use the dynamic module compilation approach for "live data" - assigns to use in page templates.  This allows users to change the value of assigns at runtime, whether the value is a simple primitive, the result of elixir code, or even dynamically generated based on a path param.

For an example of the dynamic case, imagine we have a route `shop/products/:id`
We could create a live data assign where:
```elixir
%{format: :elixir, key: "product", value: "MyApp.Shop.Products.get_product!(id)"}
```
Then create a page with path `shop/products/123` with template:
```elixir
<h2><%= @beacon_live_data[:product].name %></h2>
<img src={@beacon_live_data[:product].img_url} />
<p>Price: <%= @beacon_live_data[:product].price %></p>
etc...
```
and we have the expected results from LiveView, but everything is defined at runtime through the LiveAdmin interface.

### Implementation

* Added a `LiveData` schema which holds the path (scoped to site)
* Added a `LiveDataAssign` schema which holds key/value + the format of the value (text or elixir)
* `LiveData` has_many `LiveDataAssign`s
* Added `Content` context functions for working at both levels
* Changed `Beacon.DataSource` behaviour to no longer care about `site` because each site will have its own module
* `page_title/2` and `meta_tags/2` callbacks are deprecated (in the aforementioned behaviour)
* Page title now accepts Solid snippets
* `Beacon.Loader` now loads a data source module per site
* `DataSourceModuleLoader` generates data source modules
* Deleted unused `beacon_web/controllers/api/{page,component}_controller.ex` and the related tests which were broken by these changes
* Deleted `beacon/data_source_test.exs` which didn't seem relevant with the new approach
* Added a bunch of new tests for context functions and validating the `DataSourceModuleLoader` outputs
* Added test fixtures, removed the mock data source module

### Data Source Module Example

Generated by the DataSourceLoader (whitespace added, maps truncated here for readability):

```elixir
defmodule :"Elixir.BeaconWeb.LiveRenderer.82CBFE60DC5B9C46C4C0BC8ADDAB657ADataSource" do
  @behaviour Beacon.DataSource.Behaviour
  require Logger

  @impl Beacon.DataSource.Behaviour
  def live_data(path, params) do
    live_data(path, params, %{})
  end

  def live_data(["forum"], var!(params), var!(data)) do
    %{} = var!(params)
    Enum.reduce([%{..., key: "boards", value: "MyApp.Forum.Boards.get_board_list()"}], var!(data), fn assign, acc -> 
      Map.put(
        acc,
        String.to_atom(assign.key),
        case assign.format do
          :text -> assign.value
          :elixir -> assign.value |> Code.eval_string([], __ENV__) |> elem(0)
        end
      )
    end)
  end

  def live_data(["forum", board_slug], var!(params), var!(data)) do
    %{} = var!(params)
    Enum.reduce([%{..., key: "board", value: "MyApp.Forum.Boards.get_board_by_slug(board_slug)"}, %{..., key: "posts", value: "MyApp.Forum.Posts.get_posts_for_board(board_slug)"}], var!(data), fn assign, acc ->
      Map.put(
        acc,
        String.to_atom(assign.key),
        case assign.format do
          :text -> assign.value
          :elixir -> assign.value |> Code.eval_string([board_slug: board_slug], __ENV__) |> elem(0)
        end
      )
    end)
  end

  def live_data(["forum", board_slug, post_id], var!(params), var!(data)) do
    %{} = var!(params)
    Enum.reduce([%{..., key: "post", value: "MyApp.Forum.Posts.get_post!(post_id)"}, %{..., key: "comments", value: "MyApp.Forum.Posts.get_comments_for_post(post_id)"}], var!(data), fn assign, acc ->
      Map.put(
        acc,
        String.to_atom(assign.key),
        case assign.format do
          :text -> assign.value
          :elixir -> assign.value |> Code.eval_string([board_slug: board_slug, post_id: post_id], __ENV__) |> elem(0)
        end
      )
    end)
  end

  def live_data(path, params, data) do
    Logger.warning("Unhandled Beacon Live Data request for site #{:dockyard_com} with path #{inspect(path)} and params #{inspect(params)}")
    data
  end
end
```


lib/beacon/loader.ex
```diff
@@ -132,7 +134,8 @@ defmodule Beacon.Loader do
          :ok <- load_pages(site),
          :ok <- load_error_pages(site),
          :ok <- load_stylesheets(site),
-         :ok <- load_runtime_css(site) do
+         :ok <- load_runtime_css(site),
+         :ok <- load_data_source(site) do
```
leandrocp, line 138: I believe we need to move this call to before `load_pages`

lib/beacon/loader/data_source_module_loader.ex
```diff
@@ -0,0 +1,88 @@
...
+          unquote_splicing(live_data_functions)
+
+          def live_data(path, params, data) do
+            Logger.warning("Unhandled Beacon Live Data request for site #{unquote(site)} with path #{inspect(path)} and params #{inspect(params)}")
+            data
+          end
+        end
+      end
+
+    # For debugging - will print module content to the terminal
```
leandrocp, line 32: Since we debug AST very often, this block of code seems a good candidate to live in `Beacon.Utils` which is a private module used internally

lib/beacon/loader/data_source_module_loader.ex
```diff
@@ -0,0 +1,88 @@
...
+  defp live_data_fn(live_data) do
+    %{path: path, assigns: assigns} = live_data
+
+    path_list =
+      quote do
+        unquote(
+          path
+          |> String.split("/", trim: true)
+          |> Enum.map(fn
+            ":" <> param -> Macro.var(:"#{param}", :"#{path}")
```
leandrocp, line 53: Have you considered named parameters (for eg `/v:version`) and glob-like patterns? (for eg `/*page`), ie: support all patterns like `"/pages/he:page/*rest"`?

leandrocp (commented):
Hey @APB9785 that's just an initial review but it's looking good. Meta tags and page title are also tied to a path, so that looks like a good fit to add those in beacon_live_admin along with the assigns page. Thoughts?

lib/beacon/loader/data_source_module_loader.ex
APB9785: Named params are supported with the Phoenix router syntax i.e. `pages/:page_id/:rest` which gets converted to the "path list" format we used in the previous DataSource module i.e. `["pages", page_id, rest]` where `page_id` and `rest` are usable variables for `:elixir` format LiveDataAssigns

APB9785: @leandrocp I realize now that I misunderstood your question at first.  You are asking about support for additional patterns which would translate to something like:
```elixir
# /foo/bar:baz/bong
["foo", {"bar", baz}, "bong"] = path

# /foo/*rest
# (e.g. /foo/bar/baz/bong)
["foo" | rest] = path
```
I did consider the latter case, at least, but wasn't sure if we wanted to include that in the scope of this PR.  I think it would just require a bit more robust validation of the path, to avoid bad paths such as `foo/*bar/baz`.  Maybe it's good to add that anyway, since this pattern matching here isn't particularly robust (probably breaks on something like `/foo:/bar`).  I could work on this tomorrow, WDYT?

leandrocp:
❤️

PR #368, opened 2023-10-28
Fix Template Beacon Seeds File syntax

edborsa:
Hey @everyone.

I was just playing with the installer and I noticed a difference from the beacon_demo project and the default beacon project.
Whenever I run the seeds the url /blog/:blog_slug does not seem to work.
The reason I suspect is because the syntax used.
With the update using the syntax from inside the beacon_demo project the path seem to work.

Expected behavior:
/blog/:blog_slug should render page with matching path


leandrocp:
Thanks @edborsa

PR #376, opened 2023-11-21
Fix tailwind parser

### LINKED ISSUE #194: tailwind compiler - parse source from loaded templates
The tailwind compiler is fetching templates from DB as source to parse and find used CSS classes, which won't work with Markdown that's transformed _after_ that process. We need to change the tailwind compiler to parse from loaded templates stored on ETS, which for Markdown will be the generated HTML with all CSS classes in place.

leandrocp:
Fix #194


AZholtkevych:
Thanks a lot, @leandrocp

PR #379, opened 2023-12-01
Adds method to set an AST on a page

cibernox:
Internally it transforms that AST into a template


leandrocp:
Closing this one because we can call https://github.com/BeaconCMS/beacon/blob/d88e2f04bcc3bccd44567cf8167f4bce8faf660c/lib/beacon/template/heex/heex_decoder.ex#L20 directly since that's a public function.

Related to https://github.com/BeaconCMS/beacon_live_admin/pull/80/commits/646045eebf963d3c55e4c598e153c38ae3334d85

PR #380, opened 2023-12-01
Update page_module_loader.ex

kgautreaux:
Try to fix error in the main branch resulting from not including the page module in calls to Beacon.Template.render() when loading pages from ETS pre-rendered.


leandrocp:
Thanks!

Issue #345, opened 2023-09-07
Components and Pages: Create new layouts with a default template


leandrocp:
When a new layout is created, its template should have the following default content:

```heex
<%= @inner_content %>
```

That's necessary to render the page content using that layout, so it starts with that default content to help users creating new layouts. But it won't be enforced for now, users could remove that line.

edborsa:
Hey @leandrocp, I want to start contributing, haha.
And this one seems easy.

Here is the PR:
https://github.com/BeaconCMS/beacon/pull/383

(BTW, great talk on Elixir em Foco. I bought the "How to Open Source" book)

If there are any guidelines and tips you want to suggest, or changes, please let me know.

Best regards

leandrocp:
Closed by https://github.com/BeaconCMS/beacon/pull/383

PR #383, opened 2023-12-09
Feat: Add default value for template in layout schema

edborsa:
Issue - https://github.com/BeaconCMS/beacon/issues/345

Added default value to template attr for layout schema


leandrocp:
Thanks!

Issue #333, opened 2023-08-17
`beacon.install`: Don't allow sites starting with `beacon_`


APB9785:
The installer will allow this invalid site name, injecting it into your app, but then it will cause failures when you try to setup and build the app.

leandrocp:
See https://github.com/BeaconCMS/beacon/blob/main/lib/mix/tasks/install.ex

edborsa:
https://github.com/BeaconCMS/beacon/pull/385/files

leandrocp:
Closed by https://github.com/BeaconCMS/beacon/issues/333

PR #385, opened 2023-12-13
Issue 333 don't allow sites starting with beacon 

edborsa:
Issue: https://github.com/BeaconCMS/beacon/issues/333


leandrocp:
Nice thanks!

Issue #306, opened 2023-08-03
Route and Resource Loading: Add `--path` to `beacon.install` generator


leandrocp:
The `beacon.install` mix task accepts a `--site` option which generates the following scope in the router:

```elixir
  scope "/" do
    pipe_through :browser
    beacon_site "/my_site", site: :my_site
  end
```

We need to add a new optional `--path` option that allows changing the first argument of `beacon_site`, example:

```sh
mix beacon.install --site my_site --path /
```

Would generate:

```elixir
  scope "/" do
    pipe_through :browser
    beacon_site "/", site: :my_site
  end
```

If not specified it takes the value from `--site`.

a8t:
Still looking for help on this?

leandrocp:
> Still looking for help on this?

Yep. A PR is welcome!

edborsa:
https://github.com/BeaconCMS/beacon/pull/387

leandrocp:
Closed by #387 

/cc @AZholtkevych

PR #387, opened 2023-12-15
Feat: Route and Resource Loading Add --path to beacon.install generator

edborsa:
https://github.com/BeaconCMS/beacon/issues/306

### DIFF
--- a/lib/beacon/types/site.ex
+++ b/lib/beacon/types/site.ex
@@ -27,6 +27,8 @@ defmodule Beacon.Types.Site do
   def valid?(site) when site in ["", nil, true, false], do: false
 
   def valid?(site) when is_binary(site) do
+    # TODO: Should we add the Regex.match? from the line bellow?
+    nil
   end
 
   def valid?(site) when is_atom(site) do
@@ -42,6 +44,14 @@ defmodule Beacon.Types.Site do
 
   def valid_name?(site) when is_binary(site) do
     not String.starts_with?(site, "beacon_")
+  end
+
+  def valid_path?(path) when is_atom(path) do
+    valid_name?(Atom.to_string(path))
+  end
+
+  def valid_path?(path) when is_binary(path) do
+    String.starts_with?(path, "/")
   end
 
   @doc false
--- a/lib/beacon/types/site.ex
+++ b/lib/beacon/types/site.ex
@@ -46,7 +46,7 @@ defmodule Beacon.Types.Site do
   end
 
   def valid_path?(path) when is_atom(path) do
-    valid_name?(Atom.to_string(path))
+    path |> Atom.to_string() |> valid_path?()
   end
 
   def valid_path?(path) when is_binary(path) do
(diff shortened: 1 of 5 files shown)


lib/beacon/types/site.ex
```diff
@@ -27,6 +27,8 @@ defmodule Beacon.Types.Site do
   def valid?(site) when site in ["", nil, true, false], do: false
 
   def valid?(site) when is_binary(site) do
+    # TODO: Should we add the Regex.match? from the line bellow?
```
leandrocp, line 30: > Should we add the Regex.match? from the line bellow?

Yes.

lib/beacon/types/site.ex
```diff
@@ -44,6 +46,14 @@ defmodule Beacon.Types.Site do
     not String.starts_with?(site, "beacon_")
   end
 
+  def valid_path?(path) when is_atom(path) do
+    valid_name?(Atom.to_string(path))
```
leandrocp, line 50:
```suggestion
    path |> Atom.to_string() |> valid_name?()
```

lib/beacon/types/site.ex
```diff
@@ -44,6 +46,14 @@ defmodule Beacon.Types.Site do
...
+  end
+
+  def valid_path?(path) when is_binary(path) do
+    String.starts_with?(path, "/")
```
leandrocp, line 54: 👍🏻

lib/mix/tasks/install.ex
```diff
@@ -334,12 +342,30 @@ defmodule Mix.Tasks.Beacon.Install do
   end
 
   defp validate_options!(options) do
-    site = String.to_atom(options[:site])
+    site = options |> Keyword.get(:site, "false") |> String.to_atom()
+    path = options |> Keyword.get(:path, "/#{site}")
```
leandrocp, line 346: Actually you can just use the value from `options` directly and let `Beacon.Types.Site` do the actual validation.

```suggestion
```

And then:

```elixir
!Beacon.Types.Site.valid?(options[:site]) ...
```

leandrocp (changes requested):
Just a few comments. Thanks

lib/beacon/types/site.ex
edborsa: https://github.com/BeaconCMS/beacon/pull/387/commits/32737129a17f34cce212c5af479936281859c348

lib/beacon/types/site.ex
edborsa: https://github.com/BeaconCMS/beacon/pull/387/commits/d7eeb81b6d9efae364809cde9908932dd530f867

lib/mix/tasks/install.ex
edborsa: https://github.com/BeaconCMS/beacon/pull/387/commits/162a51ad319d3ed83e7741be375e188e849738ed

leandrocp (approved):

leandrocp:
❤️

Issue #357, opened 2023-10-10
Page Builder: Featured pages component


leandrocp:
Currently the featured blog posts section in the dockyard.com website (screenshot below) is managed by a custom database table and custom code but we want to expose that kind of content as a reusable component that will look like:

```heex
<BeaconWeb.Components.featured_pages />
```

That component can receive an `source` argument that is a function used to return a list of pages. It should be optional and the default implementation should be `Content.list_pages(site, per_page: 3, sort: :updated_at)`

`site` can be fetched from the live view process, see https://github.com/BeaconCMS/beacon/blob/cbe3bf7445cc559c96310593105e710366ed2cea/lib/beacon_web/components/components.ex#L28

<img width="1581" alt="Screenshot 2023-10-10 at 10 30 56 AM" src="https://github.com/BeaconCMS/beacon/assets/36407/00a116d7-30fe-41b0-8897-097fc1b8cde5">

edborsa:
Hey, I created a PR for this one ;)

https://github.com/BeaconCMS/beacon/pull/388

PR #389, opened 2023-12-30
Feat: Page Builder Featured Pages Component

edborsa:
Issue: https://github.com/BeaconCMS/beacon/issues/357

### DIFF
--- a/lib/beacon_web/components/components.ex
+++ b/lib/beacon_web/components/components.ex
@@ -53,6 +53,83 @@ defmodule BeaconWeb.Components do
   end
 
   @doc """
+  Renders a feature page item to be used in the BeaconWeb.Components.featured_pages component.
+
+  ## Examples
+
+      <BeaconWeb.Components.feature_page_item />
+  """
+
+  attr :title, :string, required: true
+  attr :published_date, :any, required: true
+  attr :image_path, :string, required: true
+
+  def feature_page_item(assigns) do
+    ~H"""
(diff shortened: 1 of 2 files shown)


edborsa:
Hey @leandrocp could you help me with something here?

I am not really sure what to do with @published_date and @@image_path.
From the list returned by Beacon.Content.list_pages the inserted_at is a DateTime and there is no image_path.
Another thing that I was not sure how to make it right is a missing part for the @site_pages. Is the attr attribute correct? How woul you do it?

leandrocp:
@edborsa I just realized the issue isn't well described but you're still on the right path. Components are supposed to be reusable, and even though this one is extracted from dockyard.com, it still needs to be as generic as possible providing a default implementation that can be overwritten, thus we need to:

1. let the user pass a function to fetch the pages (providing a default implementation)
2. let the user define how such pages are displayed (providing a default implementation)

So the component call will look like this:

```elixir
<BeaconWeb.Components.featured_pages />
```

Which should render a row with 3 pages displaying the `page.title` and `page.updated_at`, a link to `page.path` around each "card" and format the date as the image below:

featured pages


But the user may want to change which pages are displayed, so that component can accept a `pages` attr:

```elixir
<BeaconWeb.Components.featured_pages pages={Beacon.Content.list_pages(@beacon.site, per_page: 3, query: "products", sort: :title)} />
```

Not only that but the user may want to customize how the pages are displayed as well:

```elixir
<BeaconWeb.Components.featured_pages :let={page} pages={Beacon.Content.list_pages(...)}>
  <article ...>
    <%= page.title %>
  </article>
</BeaconWeb.Components.featured_pages>
```

---------------

Some notes:
- There is no image, that's an extra attribute used solely in the dockyard.com site
- Neither the header nor the footer has to be present in the component

---------------

I hope that's more clear now but feel free to ask. Thanks!

edborsa:
@leandrocp I think I got it.
You can use with or without inner_content.

Without it will look like this:
image

leandrocp:
❤️

PR #390, opened 2024-01-08
New Page - include default meta tags

leandrocp:
Implemented in https://github.com/BeaconCMS/beacon/commit/62094e85ee59db598909b3df37a80e7d4ec82684 but the code was incorrectly removed during the Context API refactor


APB9785 (approved):
Thanks!

Issue #403, opened 2024-02-05
Issue Compiling With `phoenix_html` Version `4.0.0`


kyleboe:
## Env
- Elixir Version (`elixir -v`): `Elixir 1.16.0 (compiled with Erlang/OTP 26)`
- Phoenix Version (`mix phx.new --version`): `1.7.11`
- Platform: macOS `14.3 (23D56)`

## Issue
Running into problems running `mix beacon.install --site my_site` following the steps in the [installation guide](https://github.com/BeaconCMS/beacon/blob/main/guides/introduction/installation.md).

```shell
$ mix beacon.install --site my_site
==> beacon
Compiling 96 files (.ex)

== Compilation error in file lib/beacon_web/live/page_live.ex ==
** (RuntimeError) use Phoenix.HTML is no longer supported in v4.0.

To keep compatibility with previous versions, add {:phoenix_html_helpers, "~> 1.0"} to your mix.exs deps
and then, instead of "use Phoenix.HTML", you might:

    import Phoenix.HTML
    import Phoenix.HTML.Form
    use PhoenixHTMLHelpers


    (phoenix_html 4.0.0) expanding macro: Phoenix.HTML.__using__/1
    lib/beacon_web/live/page_live.ex:3: BeaconWeb.PageLive (module)
    (elixir 1.16.0) expanding macro: Kernel.use/1
    lib/beacon_web/live/page_live.ex:3: BeaconWeb.PageLive (module)
could not compile dependency :beacon, "mix compile" failed. Errors may have been logged above. You can recompile this dependency with "mix deps.compile beacon --force", update it with "mix deps.update beacon" or clean it with "mix deps.clean beacon"
```

leandrocp:
Hey @kyleboe good catch, we need to support `phoenix_html ~> 2.14.2 or ~> 3.0 or ~> 4.0` and replace calls to deprecated helpers with components (the ones removed from v4.0 like `text_input`). A PR is welcome :)

PR #404, opened 2024-02-05
Bump Deps To Show Issue

kyleboe:
`mix deps.update --all` and update elixir github workflow to highlight issue mentioned in #403


leandrocp:
Hey @kyleboe thanks for the PR, that's the path to go but we also need to apply some other changes as described here https://github.com/BeaconCMS/beacon/issues/403#issuecomment-1930787128

PR #411, opened 2024-02-12
Support phoenix html 4.x

mfeckie:
I _think_ this resolves the issue reported here https://github.com/BeaconCMS/beacon/issues/403

All the tests are passing after removing the `import` and `use` of `Phoenix.HTML`.

I updated the tests to consume `input` and `button` from `BeaconWeb.CoreComponents`.

Not sure if I'm missing something important though!

### DIFF
--- a/guides/introduction/installation.md
+++ b/guides/introduction/installation.md
@@ -268,7 +268,7 @@ For more info on site options, check out `Beacon.start_link/1`.
         </ul>
 
         <.form :let={f} for={%{}} as={:greeting} phx-submit="hello">
-          Name: <%= text_input f, :name %> <%= submit "Hello" %>
+          Name: <.input type="text" field={f[:name]} /> <.button>Hello</.button>
         </.form>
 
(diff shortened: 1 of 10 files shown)


leandrocp:
❤️

leandrocp:
Hey @mfeckie had to [revert](https://github.com/BeaconCMS/beacon/pull/414) this one because we have sites running with the old Phoenix.HTML version which will require a migration before we can support v4.x

https://hex.pm/packages/phoenix_html_helpers doesn't help either because it depends on phoenix_html v4

mfeckie:
Anything I can do to help?

leandrocp:
hey @mfeckie unfortunately we're still blocked because dockyard.com is using the old `Phoenix.HTML` api and we can't use phoenix_html_helpers either, it's either the new API or the old one. So I'll work to migrate dockyard.com first then I can put this PR back into beacon main. I'll get back into this issue next week.

PR #423, opened 2024-02-29
User defined function components

### LINKED ISSUE #84: User-defined function components
We want to allow users to define their own [function components](https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html). That will ~_deprecate_~ remove the existing `my_component` feature, see https://github.com/BeaconCMS/beacon/issues/84#issuecomment-1363221968

The first step is to define such function components:

Similar to the `CoreComponents` provided by Phoenix, Beacon will provide a `UserComponents` to hold user-defined components that can be created as:

```elixir
Beacon.Components.create_component!(%{
  site: "my_site",
  name: "greet",
  attrs: [
    %{name: "name", type: "string", required: true}
  ],
  body: """
  <p>Hello, <%= @name %>!</p>
  """
})
```

Which internally compiles to:

```elixir
defmodule BeaconWeb.LiveRenderer.UserComponents{site_hash} do
  use Phoenix.Component
  
  attr :name, :string, required: true
  def greet(assigns) do
    ~H"""
    <p>Hello, <%= @name %>!</p>
    """
  end
end
```

And then one can use it in layouts and pages:

```elixir
Beacon.Pages.create_page!(%{
  path: "home",
  site: "my_site",
  layout_id: layout_id,
  template: """
  <main>
    <BeaconWeb.Components.dynamic component={"greet"} name="Jane" />
  </main>
  """
})
```

Notes:
- Name collision may happen by sharing the same module to hold all components. We could explore custom namespaces but keeping it simple for now.

leandrocp:
Closes the backend for #84


leandrocp:
@alexandrexaviersm merging this one now but feel free to open a new PR to add functionality required by live admin

PR #434, opened 2024-03-20
Redesign resource loading

leandrocp:
Previously the `Loader` module was a genserver responsible for many tasks: compile, load/reload modules, load initial modules on site init, react to events, and so on. And each "module loader" as `PageModuleLoader` would build the ast and also load (compile and load) the module. Those tasks were separated with better error handling and caching to save resources:

### Main changes
- `Content` functions `get_published_page/layout` are cached
- `Loader.Page/Layout/etc` only builds and return the quoted module (do not attempt to load it as before)
- `Compiler` does compile and load quoted forms but only if the content has changed
- `SiteSupervisor` bootstraps initial data and modules including the most recent pages
- `Loader` spawns a worker to build, compile, and load each module in an isolated process - essentially coordinates the other modules
- Replace some Content events with the general `:content_updated`. The events like `{resource}_created` and `{resource}_updated` are too granular for our use case
- `Boot` genserver replaces `Loader.load_site_from_db/1`

### Changes on ETS tables
- starts one `router` table per site instead of a global router and replace the page metadata with page_id
- introduced the `modules` table to store module name and md5 calculated from the quoted form before compilation, that's used to check if a module is fresh and skip compilation which reduces drastically that compilation error when redefining modules
- introduced the `resources` table to store published and layout pages on the `Content` api, that's because we fetch that resource during runtime, especially the page schema, so it reduces a few ms during request and also reduces DB load


lib/beacon/compiler.ex
```diff
@@ -0,0 +1,125 @@
...
+      {:ok, module, []}
+    rescue
+      error -> {:error, error, []}
+    end
+  end
+
+  def unload(module) do
+    :code.purge(module)
+    :code.delete(module)
+  end
```
APB9785, line 109: I've been wondering about the order of these two calls, ever since I tried running some benchmarks last week and reached an error which seemed to be due to these being called in the wrong order.

Looking at the docs, it says that when calling `:code.delete/1` "the current code for Module is made old"

Then `:code.purge/1` "removes code marked as old"

so it seems like it should be delete first, then purge.

There is this note in `:code.delete/1` "Returns [...] false if there is old code for Module that must be purged first" but I think this just means you can't call `:code.delete/1` twice in a row and have TWO modules worth of "old" code.  If we always call the two together, we shouldn't be affected by that.

leandrocp: It's basically swapping the current module for the new one being compiled, think of it like pointers. It works similar to mix task compile or phoenix code reloader.

First it removes (purge) the old module to make room for another one, so calling delete will move the current to old, and then compile will place the new module into current.

That works because we know a new module will be placed into current, ending up with old and current in the right spots.

So the workflow for a module `A` that gets updated from v2 to v3 would look like:

- `purge` v1 from old
- move v2 to old (delete)
- place v3 as current (compile)

And the workflow from v1 to v2, ie: in this case there's no old:

- `purge` nothing from old (returns false because there's only current v1 loaded)
- move v1 to old (delete)
- place v2 as current (compile)

It's important to keep the "previous version" as old because some process may still be using that module.

leandrocp: `delete` + `purge` is useful when you want to get rid of that module completely, so you delete first to move it to old then purge to remove the old version, ending up with nothing, the module is no longer available in any way.

lib/beacon/utils.ex
```diff
@@ -6,8 +6,11 @@ defmodule Beacon.Utils do
     Enum.reduce(list, &{:|, [], [&1, &2]})
   end
 
-  # For debugging - will print module content to the terminal
-  def ast_to_binary(ast) do
+  # For debugging - convert a quoted expression to string
+  # useful to log or write a file, eg:
+  #
+  #  File.write!("module.ex", Beacon.Utils.quoted_to_binary(quoted))
+  def quoted_to_binary(ast) do
```
APB9785, line 13: maybe we could also include an example for
```elixir
quoted |> Beacon.Utils.quoted_to_binary() |> IO.puts()
```
that's how I was mostly using it

lib/beacon_web/data_source.ex
```diff
@@ -4,30 +4,20 @@ defmodule BeaconWeb.DataSource do
...
     page =
-      assigns.__dynamic_page_id__
-      |> Beacon.Loader.page_module_for_site()
-      |> Beacon.Loader.call_function_with_retry!(:page_assigns, [])
+      site
+      |> Beacon.Loader.fetch_page_module(page_id)
+      |> Beacon.apply_mfa(:page_assigns, [])
 
-    title = BeaconWeb.Layouts.page_title(assigns)
+    title = Map.get(page, :title) || missing_page_title()
```
APB9785, line 18: Can we use `Map.fetch!/2` here instead?  At first glance, it's confusing whether the `nil` is coming from missing `:title` key or if key is present `%{title: nil}`.  But I'm thinking this is a struct so the key should always be present, right?  If so, using `fetch!` seems like it would make that explicit.  WDYT?

lib/beacon/pub_sub.ex
```diff
@@ -1,39 +1,36 @@
 defmodule Beacon.PubSub do
   @moduledoc false
 
-  alias Beacon.Content.Component
-  alias Beacon.Content.ErrorPage
-  alias Beacon.Content.Layout
-  alias Beacon.Content.LiveData
-  alias Beacon.Content.LiveDataAssign
-  alias Beacon.Content.Page
+  alias Beacon.Content
```
APB9785, line 4: I'm not a big fan of partial aliases, as they make it harder to grep the codebase for specific modules.  In here we don't even use the `Beacon.Content` module, so it also gives a different expectation for the module's dependencies at a glance.

If we used
```elixir
alias Beacon.Content.Layout
alias Beacon.Content.Page
```
that should be all we need for full aliasing here after the old code removal

leandrocp: > I'm not a big fan of partial aliases, as they make it harder to grep the codebase for specific modules.  In here we don't even use the `Beacon.Content` module, so it also gives a different expectation for the module's dependencies at a glance.
> 
> 
> 
> If we used
> 
> ```elixir
> 
> alias Beacon.Content.Layout
> 
> alias Beacon.Content.Page
> 
> ```
> 
> that should be all we need for full aliasing here after the old code removal

That's a good point, I'm gonna change it before merging.

lib/beacon_web/data_source.ex
leandrocp: > Can we use `Map.fetch!/2` here instead?  At first glance, it's confusing whether the `nil` is coming from missing `:title` key or if key is present `%{title: nil}`.  But I'm thinking this is a struct so the key should always be present, right?  If so, using `fetch!` seems like it would make that explicit.  WDYT?

Yep that field should be present so fetch! is the right choice here

lib/beacon/loader/snippets.ex
```diff
@@ -1,18 +1,21 @@
...
-    module_name = Loader.snippet_helpers_module_for_site(site)
-    functions = Enum.map(helpers, &helper/1)
-    ast = render(module_name, functions)
-    :ok = Loader.reload_module!(module_name, ast)
-    {:ok, ast}
+  defp render(module_name, [] = _functions) do
+    quote do
+      defmodule unquote(module_name) do
+      end
+    end
```
APB9785, line 18: Interesting!  What happens when we compile `(unquote_splicing([]))`?

lib/beacon/compiler.ex
APB9785: Ah, thanks so much for detailing these workflows and with your second comment I now see the difference between Beacon (keeping old code for concurrent processes) and my other case (no concurrent processes) 👍

lib/beacon/loader/snippets.ex
leandrocp: It returns an empty block but I think that's still valid so we could have a single render/2 here

lib/beacon_web/data_source.ex
APB9785: or maybe `page.title`?

APB9785 (approved):
Left a few comments but everything is looking great and runs fine locally

PR #442, opened 2024-03-29
Validate Page Event Handler code

APB9785:
This PR makes some small changes to the way elixir code validation works.  Instead of passing in the `path`, which is specifically useful for `LiveData`, now the function expects `vars` and `imports` which are decided at the callsite.  So for `LiveData`, we can parse the path and include `params`, while `PageEventHandlers` use `socket` and `event_params` and `import Phoenix.Socket`.  Both can share the same validation logic.


leandrocp (approved):
💯

PR #492, opened 2024-05-14
Introduce `@beacon` and deprecate `@beacon_*` assigns

### LINKED ISSUE #369: Validate existing beacon live data assign
Currently calling `@beacon_live_data` with an invalid key will just return `nil` since that's a regular map. That's safe but easy to miss and may result on incorrect behavior.

We want to explore ways to make `@beacon_live_data` behave more like LV `@assign` and warn and/or fail to compile if such value is not existing.

leandrocp:
Close #369 

- `@beacon` assign is a container for public and private assigns, including path params, query params, and others that used to be defined on its own assigns as `@beacon_*` or as `__{name}__` 
- `@beacon_live_data` is deprecated if favor of accessing those values from `assigns` directly, ie: `@name` instead of `@beacon_live_data[:name]`
- Add custom `assign` functions to warn and block assigning the `:beacon` key, those functions are called in event handlers and any exposed code block in the page editor

### DIFF
--- a/lib/beacon/beacon.ex
+++ b/lib/beacon/beacon.ex
@@ -194,8 +194,6 @@ defmodule Beacon do
 
     #{context}
 
-    Make sure you have created a page for this path.
-    See Pages.create_page!/2 for more info.
     """
   end
 
--- a/dev.exs
+++ b/dev.exs
@@ -256,12 +256,12 @@ dev_seeds = fn ->
         </div>
 
         <div>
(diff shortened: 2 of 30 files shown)


lib/beacon/beacon.ex
```diff
@@ -194,8 +194,6 @@ defmodule Beacon do
 
     #{context}
 
-    Make sure you have created a page for this path.
```
leandrocp, line 197: Removed because this function is reused not only for call inside pages so this phrase may get out of context

PR #520, opened 2024-06-14
Add :repo option to Beacon.Config

APB9785:
This moves us from forcing `Beacon.Repo` everywhere, to allowing the user to configure their own repo via Beacon.Config.

Much inspiration was taken from Oban, which uses the same approach

Installing Beacon is now a matter of
```
$ mix ecto.gen.migration create_beacon_tables
```
and
```elixir
defmodule MYApp.Repo.Migrations.CreateBeaconTables do
  use Ecto.Migration

  defdelegate up, to: Beacon.Migration

  defdelegate down, to: Beacon.Migration
end
```


.credo.exs
```diff
@@ -118,7 +118,7 @@
           #
           {Credo.Check.Refactor.Apply, []},
           {Credo.Check.Refactor.CondStatements, []},
-          {Credo.Check.Refactor.CyclomaticComplexity, [max_complexity: 11]},
+          {Credo.Check.Refactor.CyclomaticComplexity, [max_complexity: 12]},
```
leandrocp, line 121: No longer 🤘🏻

lib/beacon/migration.ex
leandrocp, line 1: In Oban we can call `Oban.Migration.up(version: 12)` and `Oban.Migration.down(version: 12)`. Are you thinking about introducing the same API as `Beacon.Migration`?

config/config.exs
```diff
@@ -1,12 +1,20 @@
 import Config
 
-config :beacon, ecto_repos: [Beacon.Repo]
+config :beacon, ecto_repos: [Beacon.BeaconTest.Repo]
```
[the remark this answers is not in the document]
APB9785: I think in Oban, there is no `dev.exs` or `test.exs` because all the config is only for local testing.  This was probably why I put it in `config.exs` in hopes that we could also move towards that pattern?

lib/beacon/migration.ex
APB9785: Yes but since there's only one version right now it's just a simple up/down with no args.  Eventually when there's multiple versions, I'm thinking each version has its own Module and the top-level module is a switcher based on `:version` arg just like Oban.  We could split it out now if you think it will make it easier

config/config.exs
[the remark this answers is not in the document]
leandrocp: Makes sense. Better have one file than two or more.

leandrocp (approved):
🎉

PR #522, opened 2024-06-14
Page Builder - enable the editor on New pages

ChristianTovar:
# Description

The following changes are intended for enabling the visual editor for pages that have not been published yet, therefore that do not exist on the database yet.

### DIFF
--- a/lib/beacon_web/beacon_assigns.ex
+++ b/lib/beacon_web/beacon_assigns.ex
@@ -28,22 +28,25 @@ defmodule BeaconWeb.BeaconAssigns do
             }
 
   @doc false
-  def build(site) when is_atom(site), do: %__MODULE__{site: site}
+  def build(%{site: site} = page) when is_atom(site), do: %__MODULE__{site: site, page: page}
 
   def build(beacon_assigns = %__MODULE__{}, path_info, query_params) when is_list(path_info) and is_map(query_params) do
     site = beacon_assigns.site
-    %{site: ^site} = page = Beacon.RouterServer.lookup_page!(site, path_info)
+    page = beacon_assigns.page
+
+    # TODO: The page doesn't exist yet, so we need a way to access the page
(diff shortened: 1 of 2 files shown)


lib/beacon_web/beacon_assigns.ex
```diff
@@ -28,22 +28,25 @@ defmodule BeaconWeb.BeaconAssigns do
...
-  def build(site) when is_atom(site), do: %__MODULE__{site: site}
+  def build(%{site: site} = page) when is_atom(site), do: %__MODULE__{site: site, page: page}
 
   def build(beacon_assigns = %__MODULE__{}, path_info, query_params) when is_list(path_info) and is_map(query_params) do
     site = beacon_assigns.site
-    %{site: ^site} = page = Beacon.RouterServer.lookup_page!(site, path_info)
+    page = beacon_assigns.page
+
+    # TODO: The page doesn't exist yet, so we need a way to access the page
+    # %{site: ^site} = page = Beacon.RouterServer.lookup_page!(site, path_info)
```
ChristianTovar, line 38: These changes were made in order to understand the main blocker. Normally for the visual builder we rely on a `:path` for its corresponding `:site`. That's the usual way of "fetching" a page, but since the page doesn't exist yet for new pages, we'll have to figure out a way to re adjust the function head and fix the logic so that the `lookup_page!/2` function doesn't raise.

- I'm thinking that perhaps a way to address this could be to create a page on the fly when hitting the `New page` button, that way we can have all of the requirements fullfilled, even if it is a draft.

leandrocp:
Hey @ChristianTovar I've pushed changes on main to allow building `%BeaconAssigns` with non-published pages and also non-persisted pages because we found another issue that was blocking the usage of the page builder.

https://github.com/BeaconCMS/beacon/pull/529

https://github.com/BeaconCMS/beacon/commit/b9da4b79d5dc6b607ba0ccdb5bb730bebc34e7d1

So I think we can close this PR. Thanks for the initial work!

PR #535, opened 2024-06-27
Add adjusts for the User defined Components

alexandrexaviersm:
Necessary adjusts that came up after doing more work in the UI:
- Create a new field in `beacon_component_attrs` to save the struct name of a attribute of type `struct`
- Create new context functions for `ComponentAttr`


lib/beacon/content.ex
```diff
@@ -1613,10 +1613,103 @@ defmodule Beacon.Content do
...
+
+  """
+  @spec create_component_attr(map()) :: {:ok, ComponentAttr.t()} | {:error, Changeset.t()}
+  @doc type: :component_attrs
+  def create_component_attr(attrs \\ %{}) do
+    %ComponentAttr{}
+    |> ComponentAttr.changeset(attrs)
+    |> Repo.insert()
+
+    # Do I need to do this for component_attr?
```
alexandrexaviersm, line 1636: @leandrocp

priv/repo/migrations/20240621085408_add_struct_name_to_beacon_component_attrs.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Beacon.Repo.Migrations.AddStructNameToBeaconComponentAttrs do
+  use Ecto.Migration
+
+  def change do
+    alter table(:beacon_component_attrs) do
+      add :struct_name, :string
```
alexandrexaviersm, line 6: Having this field really helped setting up the form in the UI - with this a lot of code was cut 🔪  - also, I think that is more clear to have the field `type` just to save the `component_type` and this field specific to save the **struct name** if the `type` is `struct`

lib/beacon/content.ex
```diff
@@ -1640,14 +1642,160 @@ defmodule Beacon.Content do
...
+    with {:ok, %ComponentSlot{}} <- Repo.delete(slot),
+         %Component{} = component <- Repo.preload(component, [slots: [:attrs]], force: true) do
+      {:ok, component}
+    end
+  end
+
+  @doc """
+  Add a changeset error if the option value doesn't matches the type.
+  """
+  @spec validate_if_value_matches_type(Changeset.t(), String.t(), any(), atom()) :: Changeset.t()
```
leandrocp, line 1719:
```suggestion
  @doc false
```

You can mark this function semi-private since it's not meant to be called by users, it's only internal.

test/support/migrations/20240621085408_add_struct_name_to_beacon_component_attrs.exs
leandrocp, line 1: Migrations moved to https://github.com/BeaconCMS/beacon/tree/main/lib/beacon/migrations so you can create v002 with both `alter table` statements.

_The migration in test/support/migrations is used only to execute the migrations in lib/beacon/migrations for the test env._

leandrocp (changes requested):
Looking good. Just a few comments

alexandrexaviersm:
@leandrocp it looks like `Beacon.Migration` was only running the migration `v001`. So I made some adjusts to it

leandrocp:
> @leandrocp it looks like `Beacon.Migration` was only running the migration `v001`. So I made some adjusts to it

Sorry I forgot to mention you have to bump the version here https://github.com/BeaconCMS/beacon/blob/ffbfb3d80574e7528d3aa47f5e8eda90813c8d0b/lib/beacon/migration.ex#L6

alexandrexaviersm:
> > @leandrocp it looks like `Beacon.Migration` was only running the migration `v001`. So I made some adjusts to it
> 
> Sorry I forgot to mention you have to bump the version here
> 
> https://github.com/BeaconCMS/beacon/blob/ffbfb3d80574e7528d3aa47f5e8eda90813c8d0b/lib/beacon/migration.ex#L6

@leandrocp I did bump this version number, but after that it was only running the v002

leandrocp:
> > > @leandrocp it looks like `Beacon.Migration` was only running the migration `v001`. So I made some adjusts to it
> 
> > 
> 
> > Sorry I forgot to mention you have to bump the version here
> 
> > 
> 
> > https://github.com/BeaconCMS/beacon/blob/ffbfb3d80574e7528d3aa47f5e8eda90813c8d0b/lib/beacon/migration.ex#L6
> 
> 
> 
> @leandrocp I did bump this version number, but after that it was only running the v002

Ok let's use V001 for now.

leandrocp (approved):

Issue #538, opened 2024-07-02
Beacon install missing router


aroyer:
When running the install mix task the router is missing preventing the application from starting up.

To reproduce use the following mix task followed by mix setup. 
```
mix beacon.install --site my_site --path /
mix setup
```

```
** (Mix) Could not start application site: Site.Application.start(:normal, []) returned an error: shutdown: failed to start child: Beacon
    ** (EXIT) an exception was raised:
        ** (RuntimeError) missing required option :router
            (beacon 0.1.0-dev) lib/beacon/config.ex:376: Beacon.Config.new/1
            (beacon 0.1.0-dev) lib/beacon/beacon.ex:103: Beacon.site_child_spec/1
            (elixir 1.17.1) lib/enum.ex:1703: Enum."-map/2-lists^map/1-1-"/2
            (beacon 0.1.0-dev) lib/beacon/beacon.ex:91: Beacon.init/1
            (stdlib 5.2.3) supervisor.erl:330: :supervisor.init/1
            (stdlib 5.2.3) gen_server.erl:980: :gen_server.init_it/2
            (stdlib 5.2.3) gen_server.erl:935: :gen_server.init_it/6
            (stdlib 5.2.3) proc_lib.erl:241: :proc_lib.init_p_do_apply/3
```

aroyer:
fixed the test, if this is not desired behavior then consider removing this check (https://github.com/BeaconCMS/beacon/blob/main/lib/beacon/config.ex#L376)

leandrocp:
Fixed by https://github.com/BeaconCMS/beacon/pull/539

PR #539, opened 2024-07-02
Fix for mix beacon install missing router issue #538

aroyer:
Fix for beacon install missing :router in application

https://github.com/BeaconCMS/beacon/issues/538


leandrocp:
Thanks @aroyer !

PR #542, opened 2024-07-05
Initial work for loading Tailwind plugins in beacon_live_admin

Serabe:
This is the PR for the initial work on loading Tailwind plugins on the Visual Editor.

See comments in code.

[Companion PR in beacon live admin.](https://github.com/BeaconCMS/beacon_live_admin/pull/183)

### DIFF
--- a/assets/tailwind.base.config.js
+++ b/assets/tailwind.base.config.js
@@ -1,0 +1,7 @@
+/*
+ * Add here all the config but the content.
+ */
+module.exports = {
+  theme: {},
+  plugins: [],
+}
--- a/config/config.exs
+++ b/config/config.exs
@@ -5,18 +5,23 @@ config :beacon, :generators, binary_id: true
 config :phoenix, :json_library, Jason
 
 if Mix.env() == :dev do
-  esbuild = fn args ->
+  min_esbuild = fn args ->
     [
-      args: ~w(./js/beacon.js --bundle) ++ args,
+      args: args,
       cd: Path.expand("../assets", __DIR__),
       env: %{"NODE_PATH" => Path.expand("../deps", __DIR__)}
     ]
   end
 
+  esbuild = fn args ->
+    min_esbuild.(~w(./js/beacon.js --bundle) ++ args)
+  end
+
   config :esbuild,
     version: "0.17.18",
     cdn: esbuild.(~w(--format=iife --target=es2016 --global-name=Beacon --outfile=../priv/static/beacon.js)),
-    cdn_min: esbuild.(~w(--format=iife --target=es2016 --global-name=Beacon --minify --outfile=../priv/static/beacon.min.js))
+    cdn_min: esbuild.(~w(--format=iife --target=es2016 --global-name=Beacon --minify --outfile=../priv/static/beacon.min.js)),
+    tailwind: min_esbuild.(~w(./tailwind.base.config.js --bundle --format=esm --target=es2016 --minify --outfile=../priv/static/tailwind.config.js))
 end
 
 if config_env() == :test, do: import_config("test.exs")
--- a/lib/beacon/router.ex
+++ b/lib/beacon/router.ex
@@ -86,6 +86,7 @@ defmodule Beacon.Router do
         live_session session_name, session_opts do
           get "/beacon_assets/css-:md5", BeaconWeb.AssetsController, :css, as: :beacon_asset, assigns: %{site: opts[:site]}
           get "/beacon_assets/js:md5", BeaconWeb.AssetsController, :js, as: :beacon_asset, assigns: %{site: opts[:site]}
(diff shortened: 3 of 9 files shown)


assets/tailwind.base.config.js
```diff
@@ -0,0 +1,7 @@
+/*
+ * Add here all the config but the content.
+ */
+module.exports = {
+  theme: {},
+  plugins: [],
+}
```
Serabe, line 7: This is the config that will be shared with Beacon Live Admin.

The content property is the only one dynamic, but we don't care about that one in admin, because we are providing our own content.

config/config.exs
```diff
@@ -5,18 +5,23 @@ config :beacon, :generators, binary_id: true
...
+  esbuild = fn args ->
+    min_esbuild.(~w(./js/beacon.js --bundle) ++ args)
+  end
+
   config :esbuild,
     version: "0.17.18",
     cdn: esbuild.(~w(--format=iife --target=es2016 --global-name=Beacon --outfile=../priv/static/beacon.js)),
-    cdn_min: esbuild.(~w(--format=iife --target=es2016 --global-name=Beacon --minify --outfile=../priv/static/beacon.min.js))
+    cdn_min: esbuild.(~w(--format=iife --target=es2016 --global-name=Beacon --minify --outfile=../priv/static/beacon.min.js)),
+    tailwind: min_esbuild.(~w(./tailwind.base.config.js --bundle --format=esm --target=es2016 --minify --outfile=../priv/static/tailwind.config.js))
```
Serabe, line 24: This builds the base config and expose puts it in `priv/assets`. Bundling it makes all the dependencies be in one file so we can use it safely from admin.

lib/beacon/router.ex
```diff
@@ -86,6 +86,7 @@ defmodule Beacon.Router do
         live_session session_name, session_opts do
           get "/beacon_assets/css-:md5", BeaconWeb.AssetsController, :css, as: :beacon_asset, assigns: %{site: opts[:site]}
           get "/beacon_assets/js:md5", BeaconWeb.AssetsController, :js, as: :beacon_asset, assigns: %{site: opts[:site]}
+          get "/beacon_assets/tailwind_config:md5", BeaconWeb.AssetsController, :tailwind_config, as: :beacon_asset, assigns: %{site: opts[:site]}
```
Serabe, line 89: It'd be ideal if we could import the config from Beacon as a file from an URL, and not the current method of getting the content and doing a hack for importing it.

Something like

```js
const configPromise = import(tailwindConfigUrl);
```

lib/beacon/runtime_tailwind.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Beacon.RuntimeTailwind do
```
Serabe, line 1: I use this in some way to load the content of the file from admin, but I'm not sure what this runtimes are for nor how they are being used. This quite likely needs a cleanup.

lib/beacon/tailwind_compiler.ex
```diff
@@ -49,6 +49,10 @@ defmodule Beacon.RuntimeCSS.TailwindCompiler do
   end
 
   defp generate_tailwind_config_file(site, tmp_dir, content) do
+    # Copy tailwind base config so the generated file can use it
```
Serabe, line 52: If you take a look at the eex file, it now `require`s the base config. We need to copy it next to the generated config so it can find it.

lib/beacon_web/controllers/assets_controller.ex
```diff
@@ -32,4 +32,8 @@ defmodule BeaconWeb.AssetsController do
   defp content_and_type(_site, :js) do
     {Beacon.RuntimeJS.fetch(), "text/javascript"}
   end
+
+  defp content_and_type(_site, :tailwind_config) do
```
Serabe, line 36: So we can hopefully load the config from an URL directly.

priv/static/tailwind.config.js
```diff
@@ -0,0 +1 @@
+var t=(s,e)=>()=>(e||s((e={exports:{}}).exports,e),e.exports);var g=t((i,p)=>{p.exports={theme:{},plugins:[]}});export default g();
```
Serabe, line 1: I see the builds are being committed, so I added this one too.

priv/tailwind.config.js.eex
```diff
@@ -3,11 +3,11 @@
 // See the Tailwind configuration guide for advanced usage
 // https://tailwindcss.com/docs/configuration
 
+const baseConfig = require("./tailwind.base.config");
```
Serabe, line 6: What I was mentioning in [this other comment](https://github.com/BeaconCMS/beacon/pull/542/files#r1667019785)

leandrocp:
Closed in favor of https://github.com/BeaconCMS/beacon/pull/545

PR #550, opened 2024-07-13
Validate ComponentAttr/SlotAttr Name

alexandrexaviersm:
- Move component_attr and slot_attr changeset validations to component.ex
- Add query to order Component preloads (`attrs` and `slots`) by name (if preloads are given)
- Add changeset validations - Unique name for ComponentAttr, ComponentSlot and ComponentSlotAttr

### DIFF
--- a/lib/beacon/content/component.ex
+++ b/lib/beacon/content/component.ex
@@ -13,6 +13,7 @@ defmodule Beacon.Content.Component do
 
   use Beacon.Schema
 
+  alias Beacon.Content
   alias Beacon.Content.ComponentAttr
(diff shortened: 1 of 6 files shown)


lib/beacon/content/component.ex
```diff
@@ -50,4 +51,141 @@ defmodule Beacon.Content.Component do
   end
 
   def categories, do: @categories
+
+  def validate_if_struct_name_required(changeset) do
```
leandrocp, line 55:
```suggestion
  @doc false
  def validate_if_struct_name_required(changeset) do
```

leandrocp (approved):

PR #557, opened 2024-07-19
Update content docs

APB9785:
* Adds some basic moduledoc to replace TODOs
* Group content functions which were missing group
* Move embedded schema Page.Helper into its own module so it can be undocumented
* Replace deprecated `groups_for_functions` with `groups_for_docs`
* Remove injected `child_spec` (from GenServer/Supervisor) from docs

### DIFF
--- a/lib/errors.ex
+++ b/lib/errors.ex
@@ -1,6 +1,9 @@
 defmodule Beacon.ConfigError do
   @moduledoc """
-  TODO: doc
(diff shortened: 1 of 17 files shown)


leandrocp (approved):

lib/errors.ex
leandrocp, line 1: 🙌🏻

PR #571, opened 2024-08-02
Shared event handlers

### LINKED ISSUE #519: Shared Event Handlers
A common scenario that's not yet handled fully by Beacon is handling events emitted by components. Suppose a `newsletter_form` component that renders a form with `phx-submit="signup"`, that `signup` event will be handled by an event handler defined in the Page, that works and is the current approach we have.

But it becomes cumbersome when that component is called in a layout because now all pages using that layout must create the same event handler over and over again, duplicating data and effort to manage that event.

```
layouts:
  - newsletter_form component
  pages:
    - page a
      - event_handler signup
    - page b
      - event_handler signup
```

See how the event_handler `signup` gets duplicated.

The proposal is to introduce a *(Event Handlers** resource in the Content API and also on the Admin interface to manage shared events so whenever an event is called, it will be fetched by name and executed, globally on the site.

### Tasks
- [ ] Create `beacon_event_handlers` table and schema
- [ ] Create CRUD functions in `Beacon.Content` that will be called from BeaconLiveAdmin
- [ ] Create the Event Handlers page in the admin interface
- [ ] Migrate existing Page's event handler to the shared table
- [ ] Remove Page event handlers from the admin interface
- [ ] Remove the event handler attribute/field on the Page schema and Beacon.Content

### FAQ

* Why not LiveComponents?
- Some pages may actually share the same event but not use components, so we'd need to duplicate event handlers anyway.
- The LiveComponent model is more complex than function components that we already support which would require changes to the existing approach too.

APB9785:
Resolves #519


lib/beacon/loader/page.ex
```diff
@@ -124,10 +125,10 @@ defmodule Beacon.Loader.Page do
   end
 
   defp handle_event(page) do
-    %{site: site, event_handlers: event_handlers} = page
+    event_handlers = Content.list_event_handlers(page.site)
```
leandrocp, line 128: There's still one step missing here :)

Whenever an event handler changes we'd need to broadcast a message to recompile all page modules so I'm wondering if it would be better to create a new `EventHandler` module containing all `handle_event` and on the `handle_event` of `PageLive` here https://github.com/BeaconCMS/beacon/blob/ce916b5c5f20e94737f3f6b54e1356ae525d6be9/lib/beacon/web/live/page_live.ex#L70-L76

We just replace `page_module` with `event_handler_module`.

In this case whenever an event handler changes we only recompile the single event handler module and save resources by not duplicating the same event_handler on all page modules.

Wdyt?

PR #599, opened 2024-10-06
add guide for on_mount and handle_info

djcarpe:
After getting help in the beacon slack, I promised to add a guide for `on_mount` and `handle_info` within a given page.

The example is simple just adding a constantly updating clock.

Let me know if you have a better idea of how to show this feature and to let it shine.

Thanks!

### DIFF
--- a/guides/recipes/on_mount_handle_info_loop.md
+++ b/guides/recipes/on_mount_handle_info_loop.md
@@ -1,0 +1,282 @@
+# On Mount Handle Info Loop
+
+If you would like to continuously monitor or update a page, you can create an [on_mount](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#on_mount/1) callback, and corresponding [handle_info](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#c:handle_info/2) handler.
+
+In this example we're going to update a `DateTime` every `1_000` milliseconds from the server, starting that loop `on_mount`, then continuously looping over a `handle_info`.
+
+## Creating the On Mount Handler Module
+
+After completing [your first site](../introduction/your-first-site.md) guide, create a module to handle 
+your `on_mount` callbacks. This is an example, so find a place that works best 
+for your project structure. Here, we're creating a `my_site_on_mount.ex` file
+in our web path.
+
+```bash
+├── lib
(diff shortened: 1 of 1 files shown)


guides/recipes/on_mount_handle_info_loop.md
```diff
@@ -0,0 +1,282 @@
+# On Mount Handle Info Loop
+
+If you would like to continuously monitor or update a page, you can create an [on_mount](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#on_mount/1) callback, and corresponding [handle_info](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#c:handle_info/2) handler.
```
leandrocp, line 3: And here we could add few words to highlight some problems this pattern can solve:

```suggestion
Using [Live Data](https://hexdocs.pm/beacon/0.1.0-rc.2/Beacon.Content.html#create_live_data/1) assigns, [on_mount](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#on_mount/1) callbacks, and  corresponding [handle_info](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#c:handle_info/2) handlers is a pattern that you can leverage to achieve optimistic UI for your pages.

Suppose you're displaying a Weather widget on your page but fetching the data takes a second, delaying the whole page, even though that widget is not critical to your page and could display dummy data (or no data at all) while the rest of the page is loaded. In this scenario, a Live Data assign would provide the initial dummy data, the on_mount hook request the data to be loaded asyncly, and finally the handle_info loads the data and update that assign.

For this example, we'll build a similar scenario but also add a continuous monitor to update the assign frequently.
```

leandrocp (commented):
This guide is so useful, well done. ❤️

Just left a suggestion to make it clear that this pattern is not only useful to update the assigns but also to achieve optimistic UIs to make the pages load faster and perform better. Wdyt? Thanks!

djcarpe:
Thanks for the feedback! I committed your suggestions. I can clean up and reread this weekend or feel free to merge if you’re happy with it.

leandrocp:
Documentation is a never-ending job :)
This version is already great. Thanks!

PR #608, opened 2024-10-16
Introduce config `:tailwind_css`

leandrocp:
This is a recurring question in #beacon-cms and a fair one since Beacon and the host app share some resources. The current `stylesheet` resource provided by Beacon doesn't solve the problem because it only allows to link an external stylesheet but that means duplicating Tailwind css in the same document causing conflicts and sending unnecessary data over the wire.

So this config permits reusing the same app.css from host app to generate the stylesheet for beacon pages. The only requirement are:
- app.css must load `@tailwind base/components/utilities` but that's the default anyway
- generated (compiled) app.css must be accessible from within the release, ie: must be in the priv/ dir which is also the default config in new apps

So both are expected to be met in regular Phoenix apps.

I'll update https://github.com/BeaconCMS/beacon/pull/606 to reflect this new config.


APB9785 (approved):
looks good! 🚀

PR #621, opened 2024-10-25
Allow to patch (navigate patching the content) to another site

leandrocp:
Suppose the following routes:

```elixir
scope "/", host: "sitea.com" do
  pipe_through [:browser]
  beacon_site "/sitea", site: :site_a
end

scope "/", host: "siteb.com" do
  pipe_through [:browser]
  beacon_site "/siteb", site: :site_b
end
```

Given you're in any page of site_a and there's a link to another site such as http://siteb.com:4000/other

This change will allow Beacon to patch the contents of the page and display the /other page from site_b

That's very powerful because it allows to navigate between sites without requiring full refreshes. For instance LiveView does allow this kind of navigation too.

Note that patching to relative paths is not supported by LV and won't be supported by Beacon either. For eg a patch link to /other will raise.

Thanks to @APB9785 for the discussion on this issue.

### DIFF
--- a/lib/beacon/private.ex
+++ b/lib/beacon/private.ex
@@ -1,0 +1,35 @@
+defmodule Beacon.Private do
+  @moduledoc false
+
+  # Concentrate calls to private APIs so it's easier to track breaking changes and document them
+  # in case we need to make changes.
+  # Should be avoided as much as possible.
+
+  @doc """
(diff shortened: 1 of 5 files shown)


APB9785 (approved):
nice! 🚀

PR #644, opened 2024-11-11
fix: serve media library assets inside aliased scopes

### LINKED ISSUE #642: `beacon_site` should work in scopes with prefix
```
 warning: GremioGarraAppWeb.Beacon.Web.MediaLibraryController.init/1 is undefined (module GremioGarraAppWeb.Beacon.Web.MediaLibraryController is not available or is yet to be defined)
    │
 81 │     beacon_site "/", site: :gremio_garra
    │     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

That route should reset the nested scope so the module is resolved correctly.

leandrocp:
Close #642

Now it allows to use `beacon_site` inside aliased scopes and also doesn't expose the site name.

### DIFF
--- a/dev.exs
+++ b/dev.exs
@@ -79,7 +79,7 @@ defmodule DemoWeb.Router do
     plug :put_secure_browser_headers
   end
 
-  scope "/" do
+  scope "/", DemoWeb do
     pipe_through :browser
     beacon_site "/dev", site: :dev
     beacon_site "/dy", site: :dy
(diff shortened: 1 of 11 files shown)


matthewsinclair (approved):
FWIW, I have tested this in my example (trying to integrate Ash and Beacon) and this now allows me to have `beacon_site` calls within a paramaterised `scope` block. Thanks!

APB9785 (approved):
works for me 🥳

PR #647, opened 2024-11-12
Loader - global lock using Registry

APB9785:
Lots of noisy changes from renaming Loader functions from `reload_` to `load_` to better match the latest changes.  For example, a PubSub ping that content was updated, will not do any "reloading" but rather "unload" the old content.  The reloading will not happen right away anymore - it's not until the next request comes in that needs the new content, when it will be loaded.

The main changes are:
* Addition of `safe_load/2` helper in `Beacon.Loader.Worker` to wrap all `load_` logic
* Setting `Beacon.ErrorHandler` flag for all `Beacon.Loader.Worker` processes

### DIFF
--- a/lib/beacon/loader/worker.ex
+++ b/lib/beacon/loader/worker.ex
@@ -571,6 +571,7 @@ defmodule Beacon.Loader.Worker do
       {:ok, _pid} ->
         # we are the first worker, let's do the work
         load_fn.()
+        Registry.unregister(Beacon.Registry, module)
         module
 
       {:error, {:already_registered, pid}} ->
@@ -580,7 +581,7 @@ defmodule Beacon.Loader.Worker do
         receive do
           {:DOWN, _ref, :process, _pid, {:shutdown, :loaded}} -> module
         after
-          10_000 -> :error
+          1_000 -> :error
         end
     end
   end
--- a/lib/beacon/loader/worker.ex
+++ b/lib/beacon/loader/worker.ex
@@ -376,6 +376,12 @@ defmodule Beacon.Loader.Worker do
     |> stop(config)
   end
 
+  def handle_call({:load_page_module, page_id, module}, _from, config) do
+    ^module = load_page_module(config.site, page_id, module)
+
+    {:stop, {:shutdown, :loaded}, module, config}
+  end
+
   def handle_call(:reload_runtime_js, _from, config) do
     stop(Beacon.RuntimeJS.load!(), config)
   end
@@ -477,6 +483,39 @@ defmodule Beacon.Loader.Worker do
         result = compile_module(ast, "page")
         :ok = Beacon.PubSub.page_loaded(page)
         result
+    end
+  end
+
+  def load_page_module(site, page_id, module) do
+    page = Beacon.Content.get_published_page(site, page_id)
+
+    case page do
+      nil ->
+        {:error, :page_not_published}
+
+      page ->
+        case Registry.register(Beacon.Loader.Registry, module, module) do
+          {:ok, _pid} ->
+            ast = Beacon.Loader.Page.build_ast(site, page)
+            {:ok, ^module} = compile_module(ast, "page")
+            :ok = Beacon.PubSub.page_loaded(page)
+            module
+
+          {:error, {:already_registered, pid}} ->
+            # another worker already started, let's wait for it
+            case Process.monitor(pid) do
+              {:DOWN, _ref, :process, _pid, {:shutdown, :loaded}} ->
+                module
+
+              _ref ->
+                receive do
+                  {:DOWN, _ref, :process, _pid, {:shutdown, :loaded}} ->
+                    module
+                after
+                  10_000 -> :error
+                end
+            end
+        end
     end
   end
 
--- a/lib/beacon/loader/worker.ex
+++ b/lib/beacon/loader/worker.ex
@@ -410,7 +410,7 @@ defmodule Beacon.Loader.Worker do
   end
 
   def load_snippets_module(site) do
-    module = Loader.module_name(site, "Snippets")
+    module = Loader.fetch_snippets_module(site)
 
     safe_load(module, fn ->
       snippets = Content.list_snippet_helpers(site)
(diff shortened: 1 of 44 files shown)


lib/beacon/loader/worker.ex
```diff
@@ -523,4 +563,26 @@ defmodule Beacon.Loader.Worker do
...
+        # another worker already started, let's wait for it
+        _ref = Process.monitor(pid)
+
+        receive do
+          {:DOWN, _ref, :process, _pid, {:shutdown, :loaded}} -> module
+        after
+          1_000 -> :error
+        end
+    end
+  end
```
APB9785, line 587: This is the main logic.  Tried it without the `unregister/2` at first and it had some issues in running tests, it would take forever to finish, I think due to the sandboxing.

So now it unregisters after compiling, and also lowered the timeout to 1_000

lib/beacon/router_server.ex
```diff
@@ -23,7 +23,7 @@ defmodule Beacon.RouterServer do
     # We store routes by order and length so the most visited pages will likely be in the first rows
     :ets.new(table_name(config.site), [:ordered_set, :named_table, :public, read_concurrency: true])
 
-    if config.mode == :live do
+    if config.mode in [:live, :manual] do
```
APB9785, line 26: this was for local testing, the dev script wasn't working without `async_init` being run

lib/beacon/router_server.ex
```diff
@@ -38,20 +38,11 @@ defmodule Beacon.RouterServer do
...
+      Beacon.apply_mfa(page_module, :page, [])
     else
-      with {_path, page_id} <- lookup_path(site, path_info) do
-        page_module = Beacon.Loader.fetch_page_module(site, page_id)
-        Beacon.apply_mfa(page_module, :page, [])
-      else
-        _ -> nil
-      end
+      _ -> nil
     end
```
APB9785, line 46: this is now unified so all modes will use the error handler for loading (via apply/3)

lib/beacon/web/live/page_live.ex
```diff
@@ -20,9 +20,10 @@ defmodule Beacon.Web.PageLive do
     %{"beacon_site" => site} = session
 
     # Use Beacon custom error handler to automatically load modules on-demand
-    if Beacon.Config.fetch!(site).mode == :live do
-      Beacon.ErrorHandler.enable(site)
-      if connected?(socket), do: :ok = Beacon.PubSub.subscribe_to_page(site, path)
+    Beacon.ErrorHandler.enable(site)
+
+    if Beacon.Config.fetch!(site).mode == :live and connected?(socket) do
+      :ok = Beacon.PubSub.subscribe_to_page(site, path)
```
APB9785, line 26: always enable the error handler, no matter what mode

lib/beacon/web/live/page_live.ex
```diff
@@ -109,12 +110,15 @@ defmodule Beacon.Web.PageLive do
...
+
+        # refresh the error handler in case the site changed
+        Beacon.ErrorHandler.enable(site)
+
         page = RouterServer.lookup_page!(site, path_info)
         live_data = Beacon.Web.DataSource.live_data(site, path_info, Map.drop(params, ["path"]))
         beacon_assigns = BeaconAssigns.new(site, page, live_data, path_info, params)
 
         if socket.assigns.beacon.site != site && Beacon.Config.fetch!(site).mode == :live do
-          Process.put(:__beacon_site__, site)
```
APB9785, line 117: again, this needs to be set for automated/local testing

lib/beacon/router_server.ex
leandrocp, line 26: do we still need this change after https://github.com/BeaconCMS/beacon/pull/649?

leandrocp (commented):
🎉 
Just a few minor comments

test/beacon/lifecycle/template_test.exs
```diff
@@ -4,6 +4,18 @@ defmodule Beacon.Lifecycle.TemplateTest do
...
   alias Beacon.Lifecycle
 
+  setup do
+    site = default_site()
+
+    # we aren't passing through PageLive normally in these tests so we have to manually
+    # enable the ErrorHandler and set the site in the Process dictionary
+    # (which would normally happen in the LiveView mount)
+    Process.put(:__beacon_site__, site)
+    Process.flag(:error_handler, Beacon.ErrorHandler)
```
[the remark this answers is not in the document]
APB9785: I'd prefer to not use public Beacon functions in tests, because if the logic of `enable/1` ever changes, it could affect the results of the test in a way that was not intended.

lib/beacon/web/data_source.ex
```diff
@@ -17,22 +17,15 @@ defmodule Beacon.Web.DataSource do
...
-          end
+        {:ok, %{path: path, title: title}} = page_assigns
 
         Logger.error("""
         failed to interpolate page title variables
 
         will return the original unmodified page title
 
-        site: #{site}
+        site: #{title}
```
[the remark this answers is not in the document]
APB9785: good catch

lib/beacon/router_server.ex
APB9785: Maybe not!  I'll try without the change and see if it still works

test/beacon/lifecycle/template_test.exs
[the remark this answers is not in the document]
leandrocp: okay

lib/beacon/web/live/page_live.ex
```diff
@@ -109,12 +110,15 @@ defmodule Beacon.Web.PageLive do
 
       site ->
         %{"path" => path_info} = params
+
+        # refresh the error handler in case the site changed
+        Beacon.ErrorHandler.enable(site)
```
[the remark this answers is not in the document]
APB9785: Ah, I think I had an issue with tests failing before, which was fixed by moving `Beacon.ErrorHandler.enable/1` outside the conditional.  But I just tried again, and I think that issue was resolved in another way, because it works fine the way you wrote it.  So I'll move it back inside the conditional

PR #649, opened 2024-11-13
fix: load resources on dev.exs script

leandrocp:


dev.exs
```diff
@@ -38,7 +38,7 @@ _ = Ecto.Adapters.Postgres.storage_up(Demo.Repo.config())
 path = Path.join([Path.dirname(__ENV__.file), "test", "support", "migrations"])
 Demo.Repo.start_link()
 # TODO: script arg to skip recreating the data (to avoid losing temporary data)
-Ecto.Migrator.run(Demo.Repo, path, :down, all: true, log_migrations_sql: true)
+# Ecto.Migrator.run(Demo.Repo, path, :down, all: true, log_migrations_sql: true)
```
leandrocp, line 41: Dropping all data was necessary due to an old limitation that is no longer present, so we can upsert the data now. Useful for continuous tests after restarting dev.exs

APB9785: ah, this is what was causing the router table to be empty earlier

leandrocp:
@APB9785 this should be enough to boot sites manually so we can restart sites from `:manual` to `:live` since now we're starting all process, not only `Boot.init`

PR #659, opened 2024-11-18
Only start reachable sites

leandrocp:
Supposed the following router:

```elixir
scope "/admin", host: ["beacon-site-a.fly.dev", "beacon-site-b.fly.dev"] do
  pipe_through [:browser, :beacon_admin]
  beacon_live_admin "/"
end
  
scope "/", MyAppWeb, host: ["beacon-site-a.fly.dev"] do
  pipe_through [:browser]
  beacon_site "/a", site: :site_a
  beacon_site "/b", site: :site_b
end

scope "/", MyAppWeb, host: ["beacon-site-b.fly.dev"] do
  pipe_through [:browser]
  beacon_site "/a", site: :site_a
  beacon_site "/b", site: :site_b
end
```

Suppose you deploy beacon-site-a.fly.dev and beacon-site-b.fly.dev in two different apps so you'll have:

App **A** on domain beacon-site-a.fly.dev starting `site_a` and `site_b`
App **B** on domain beacon-site-b.fly.dev starting `site_a` and `site_b`

Each app starts the two sites because it's the same codebase but notice that no requests will ever reach `site_b` on app **A** because `:host` will only match beacon-site-a.fly.dev (that's why you can have two sites running at "/").

The first obvious problem here is that we're wasting resources starting a site that will never be used in that node, but that also brings some issues to BeaconLiveAdmin since we can't properly generate URLs for assets and other resources."

With this change Beacon won't start unreachable sites to save resources and avoid crashes due to malformed URLs.


guides/deployment/deployment-topologies.md
```diff
@@ -0,0 +1,386 @@
...
+
+The URL of a Beacon Page inside a Phoenix Router [scope](https://hexdocs.pm/phoenix/Phoenix.Router.html#scope/2)
+is generated by the combination of the Endpoint and the Router configuration as `endpoint host` + `scope path` + `site prefix` + `page path`, for example:
+
+```elixir
+# endpoint
+config :my_app, MyAppWeb.Endpoint, url: [host: "mysite.com"]
+
+# router scope
+scope "2024" do
```
APB9785, line 24: should it be `scope "/2024" do` ?

lib/beacon/router.ex
```diff
@@ -246,4 +246,39 @@ defmodule Beacon.Router do
...
+  #   scope "/", MyAppWeb, host: ["beacon-site-b.fly.dev"] do
+  #     pipe_through [:browser]
+  #     beacon_site "/", site: :site_b
+  #   end
+  #
+  # On a node deployed to beacon-site-a.fly.dev, the second `beacon_site`
+  # will never match, so starting `:site_b` is a waste of resources
+  # and a common cause of problems on BeaconLiveAdmin since we can't resolve URLs properly.
+  #
+  # Simillarly, if a `get "/"` is added _before_ either `beacon_site` that `get` would always
```
APB9785, line 269:
```suggestion
  # Similarly, if a `get "/"` is added _before_ either `beacon_site` that `get` would always
```

APB9785 (approved):

PR #672, opened 2024-11-25
Page warming

### LINKED ISSUE #666: "Page Warming"
Pages are now loaded 100% lazily.  This means the first time a page is requested, it takes longer to serve than subsequent requests - which could impact SEO.  One thing we can do to resolve this is "page warming" by choosing a few of the most popular pages and pre-loading them on boot.  Since we don't know yet what the most popular pages are, a reasonable default is to pick the ones with the shortest path (the "root" pages).  Eventually we can allow users to override the default via Config, if they know of a page with a long path, but is still an ideal candidate for warming.

APB9785:
## Resolves #666 

Adds feature to warm pages on Beacon Boot.  Configured via `Beacon.Config` `:page_warming`. Allowed options are:
* `{:shortest_paths, n}` - warms the `n` pages with the shortest paths (using database `LENGTH()` function)
* `{:specific_paths, list_of_path_strings}` - attempts to warm all pages with the provided paths.  skips if the page is unpublished
* `:none` - disables the page warming feature entirely

### DIFF
--- a/lib/beacon/boot.ex
+++ b/lib/beacon/boot.ex
@@ -34,15 +34,14 @@ defmodule Beacon.Boot do
 
     %{mode: :live} = Beacon.Config.update_value(site, :mode, :live)
 
-    assets = [
+    tasks = [
       Task.Supervisor.async(task_supervisor, fn -> Beacon.Loader.load_runtime_js(site) end),
       Task.Supervisor.async(task_supervisor, fn -> Beacon.Loader.load_runtime_css(site) end)
+      | warm_pages_async(task_supervisor, config)
     ]
 
     # TODO: revisit this timeout after we upgrade to Tailwind v4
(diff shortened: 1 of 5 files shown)


lib/beacon/boot.ex
```diff
@@ -42,9 +42,33 @@ defmodule Beacon.Boot do
     # TODO: revisit this timeout after we upgrade to Tailwind v4
     Task.await_many(assets, :timer.minutes(5))
 
+    warm_pages(config)
```
leandrocp, line 45: We can make it run async along with JS and CSS so it loads assets and warm up pages at the same time (one doesn't depend on the other).

```elixir
resources = [
  css,
  js,
  pages
]

await_many(resources)
```

leandrocp (approved):

PR #684, opened 2024-12-04
fix: forward live data errors

leandrocp:
Live Data are invoked with Beacon.Loader which would rescue and reraise causing the original error to be lost, but that should not happen for Live Data calls since users might raise custom errors.

### DIFF
--- a/lib/beacon/loader.ex
+++ b/lib/beacon/loader.ex
@@ -41,6 +41,8 @@ defmodule Beacon.Loader do
   rescue
     error ->
       if live_data_module?(module) do
+        # LiveData is user-provided code, which always has the possibility of errors.
+        # In this case, we want to ensure the original error is surfaced to the user for easier debugging.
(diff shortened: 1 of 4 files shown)


lib/beacon/loader.ex
```diff
@@ -40,13 +40,24 @@ defmodule Beacon.Loader do
     end
   rescue
     error ->
-      raise_invoke_error(site, error, module, function, args, opts[:context], __STACKTRACE__)
+      if live_data_module?(module) do
+        reraise error, __STACKTRACE__
```
APB9785, line 44:
```suggestion
      if live_data_module?(module) do
        # LiveData is user-provided code, which always has the possibility of errors.
        # In this case, we want to ensure the original error is surfaced to the user for easier debugging.
        reraise error, __STACKTRACE__
```

APB9785 (approved):

PR #687, opened 2024-12-04
Beacon generates `sitemap.xml` per site and a `sitemap_index.xml`

### LINKED ISSUE #169: Feature: generate site map
Generating a site map should come from Beacon

APB9785:
## Resolves #169

Depends on #694


lib/beacon/loader/routes.ex
```diff
@@ -49,6 +49,17 @@ defmodule Beacon.Loader.Routes do
           @endpoint.url() <> beacon_media_path(file_name)
         end
 
+        def beacon_page_url(%{path: path} = page) do
+          prefix = @router.__beacon_scoped_prefix_for_site__(@site)
```
leandrocp, line 53: Better use `Phoenix.VerifiedRoutes.unverified_path/3` here, for an example see: https://github.com/BeaconCMS/beacon/blob/ba5e3be248d942120c9842aa12c83061f0dce7c9/lib/beacon/web/components/layouts.ex#L33-L34

Same for `beacon_sitemap_url`

lib/beacon/router.ex
```diff
@@ -153,6 +156,19 @@ defmodule Beacon.Router do
     end
   end
 
+  @doc """
+  Creates a beacon sitemap index.
+  """
+  defmacro beacon_sitemap_index(filename) do
+    quote bind_quoted: binding(), location: :keep do
```
leandrocp, line 163: I think we should start using `generated: true` as well for injected code (the other macro is missing it but I think it should have it as well).

```suggestion
    quote bind_quoted: binding(), location: :keep, generated: true do
```

lib/beacon/router.ex
```diff
@@ -153,6 +156,19 @@ defmodule Beacon.Router do
...
+      import Phoenix.Router, only: [scope: 3, get: 3, get: 4]
```
leandrocp, line 164: Will `get: 3` be used here?

```suggestion
      import Phoenix.Router, only: [scope: 3, get: 4]
```

lib/beacon/router.ex
```diff
@@ -153,6 +156,19 @@ defmodule Beacon.Router do
     end
   end
 
+  @doc """
+  Creates a beacon sitemap index.
```
leandrocp, line 160: I guess you already knot that but we'll need more docs here :)
Or a recipe to show how to use sitemaps.

lib/beacon/web/controllers/sitemap_controller.ex
```diff
@@ -0,0 +1,47 @@
...
+
+  defp get_pages(site) do
+    routes_module = Beacon.Loader.fetch_routes_module(site)
+
+    site
+    |> Beacon.Content.list_published_pages()
+    |> Enum.map(fn page ->
+      %{
+        loc: Beacon.apply_mfa(site, routes_module, :beacon_page_url, [page]),
+        lastmod: page.updated_at
```
leandrocp, line 43: The spec mentions that lastmod must follow https://www.w3.org/TR/NOTE-datetime which is essentially ISO8601 so we'll need to format the datetime accordingly.

leandrocp (changes requested):
Super cool. Left some comments but that's looking good.

leandrocp (approved):

PR #694, opened 2024-12-10
Add `Beacon.ProxyEndpoint` to serve multiple domains

leandrocp:
- [ ] Proxy Endpoint
- [ ] Igniter tasks

Usage example: https://github.com/BeaconCMS/beacon_demo/pull/19


APB9785 (approved):
sitemap PR seems to work OK with the proxy endpoint

PR #698, opened 2024-12-10
fix: make Router.reachable/2 less strict

leandrocp:
Allow root paths / to coexist and start beacon sites even if there's a conflict to allow nested beacon pages as /:page to be served, even in the presence of another live / route.

This change will potentially cause more sites to be started but being to restrictive is a breaking change either, so we're partially reverting to v0.2 behavior.

### DIFF
--- a/lib/beacon/router.ex
+++ b/lib/beacon/router.ex
@@ -278,7 +278,7 @@ defmodule Beacon.Router do
         router.__beacon_scoped_prefix_for_site__(site)
       end)
 
-    case Phoenix.Router.route_info(router, "GET", prefix, host) do
+    case Phoenix.Router.route_info(router, "GET", prefix, host) |> dbg do
       # bypass and allow booting beacon sites even though there's a route conflict
       # but only for root paths, for example:
       #   live /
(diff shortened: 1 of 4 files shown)


APB9785 (commented):
could it be a problem if someone defines
```elixir
live "/:page"
beacon_site "/"
```
?

leandrocp:
> could it be a problem if someone defines
> 
> ```elixir
> live "/:page"
> beacon_site "/"
> ```
> 
> ?

The site wouldn't boot in that scenario https://github.com/BeaconCMS/beacon/pull/698/commits/35b157b0dc7dc392a48145d8c2395e2c1fcabb1c

leandrocp:
@APB9785 merging this one now because I need that code :D
but we can review/refine it tomorrow

absowoot:
This PR fixed my issue where a site would not boot because it was not first in the router, but it may have introduced a new problem. When viewing pages built with Beacon, images only load sometimes.


https://github.com/user-attachments/assets/6ac28166-e00c-4992-bdfd-0fba97a8d0d1

PR #699, opened 2024-12-10
fix: boot beacon even without defined beacon_site routes in host app …

leandrocp:
Starting a site in the sup tree is not enough, each site must have a corresponding route in the host app's router. But in some cases the route might be missing, which would lead to boot crashes.

Now we're guarding against such scenario and allowing Beacon to boot, but we emit a warning with a link to a troubleshot guide.

### DIFF
--- a/guides/general/troubleshooting.md
+++ b/guides/general/troubleshooting.md
@@ -29,9 +29,13 @@ a site prefix can never match and it will never receive requests.
 
 That's is not necessarily an error if you have multiple sites in the same project
 and each scope is filtering requests on the `:host` option.
+But it may indicate:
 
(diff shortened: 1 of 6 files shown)


APB9785:
this covers the case of zero sites in the router, but what about cases where 2 sites are started in config, but only 1 is defined in the router?  would that still cause the boot to fail?

APB9785 (approved):
ah, I see, this affects the `Enum.reduce` we do over the list of sites in `Beacon.init/1`, works for me 🚀

leandrocp:
> ah, I see, this affects the `Enum.reduce` we do over the list of sites in `Beacon.init/1`, works for me 🚀

yep although needs to verify some other edge case, so merging this one now and I will write a few more tests tomorrow

PR #700, opened 2024-12-11
Beacon generates `robots.txt` per site

### LINKED ISSUE #219: Webfiles and Metatags: Generate `robots.txt`
Generate robots.txt for sites.

Each site will have its own robots.txt which must be resolved dynamically by adding a route `/robots.txt` to https://github.com/BeaconCMS/beacon/blob/7790eb72769a026c0bdfc3167aab394a9b73ce91/lib/beacon/router.ex#L79

A request to that route should call `Beacon.Lifecycle.generate_robots_txt/1` which will provide a default implementation that should work for most scenarios:

```txt
# http://www.robotstxt.org
User-agent: *
Sitemap: #{endpoint.url()}/sitemap.xml
```

`generate_robots_txt/1` should receive `site` as argument and call `Elixir.Beacon.Config.fetch!(site).endpoint.url()` to fetch current site url to be used as prefix for sitemap.xml location.

Then that content should be served as txt.

Depends on #169 

## Refs
- http://www.robotstxt.org
- https://developers.google.com/search/docs/crawling-indexing/robots/intro

APB9785:
## Resolves #219 

Depends on #694, #687

### DIFF
--- a/lib/beacon/web/robots/robots.txt.eex
+++ b/lib/beacon/web/robots/robots.txt.eex
@@ -1,0 +1,3 @@
+# http://www.robotstxt.org
+User-agent: *
+Sitemap: <%= @sitemap_url %>
--- a/lib/beacon/router.ex
+++ b/lib/beacon/router.ex
@@ -149,6 +149,7 @@ defmodule Beacon.Router do
           get "/__beacon_assets__/js-:md5", Beacon.Web.AssetsController, :js, assigns: %{site: opts[:site]}
 
           get "/sitemap.xml", Beacon.Web.SitemapController, :show, as: :beacon_sitemap, assigns: %{site: opts[:site]}
+          get "/robots.txt", Beacon.Web.RobotsController, :show, as: :beacon_robots, assigns: %{site: opts[:site]}
 
           live "/*path", Beacon.Web.PageLive, :path
         end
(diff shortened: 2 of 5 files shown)


lib/beacon/web/robots/robots.txt.eex
```diff
@@ -0,0 +1,3 @@
+# http://www.robotstxt.org
```
leandrocp, line 1: We could disallow the admin prefix here as well. See https://github.com/BeaconCMS/beacon_live_admin/blob/f8ff1621253c30c5f143655c11acf1116fa1c1e4/lib/beacon/live_admin/router.ex#L88

We need to use `scoped_path` to build the full path including the `scope` path. So if that attribute is present in the router we could include it as `Disallow: <%= @admin_prefix %>` otherwise just ignore the line since the admin might not even be included in the router.

And we'll need at least a simple test because that function might change and we need a failing test if that happens.

leandrocp (commented):
Looking good. Just a suggestion to disallow the admin prefix.

lib/beacon/web/robots/robots.txt.eex
APB9785: > We could disallow the admin prefix here as well

I don't think that will be included.  This PR only uses the `beacon_site` macro for generation

PR #708, opened 2024-12-17
fix: load visual editor on new pages

leandrocp:
`Beacon.Content.get_page/2` fails since there's no page saved yet. This PR also simplified a few things:

- Simplify the build of a new `@beacon` assign
- Make the arguments more explicit
- Remove the `source` arg so callers don't need to be responsible for how the assign is built

### DIFF
--- a/lib/beacon/router.ex
+++ b/lib/beacon/router.ex
@@ -250,6 +250,8 @@ defmodule Beacon.Router do
     end)
   end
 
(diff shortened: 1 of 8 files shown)


APB9785 (approved):
It's a nice improvement to infer the published status from `id: nil` instead of having the caller pass in a source :rocket:

PR #713, opened 2025-01-02
Fix media uploads in multi node deployments

### LINKED ISSUE #709: Multi-node deployments fail to upload files to MediaLibrary
### Environment

* Elixir version (elixir -v): 1.17.1
* Phoenix version (mix deps): 1.7.14
* Phoenix LiveView version (mix deps): 0.20.12
* Beacon version (mix deps): 0.3.3
* Operating system: Flyio
* Browsers on which you attempted to reproduce this bug (the more the merrier): N/A

### Actual behavior

On a multi-node deployment, file uploads fail during processing
* This is due to the `/tmp` file location being on one node, while another node attempts image processing

### Expected behavior

File upload is OK

APB9785:
## Resolves #709 

UploadMetadata will now hold a `:node` along with the `:path` of the file.  This ensures that in a multi-node setup, we are looking on the appropriate node when reading the file.

### DIFF
--- a/lib/beacon/media_library.ex
+++ b/lib/beacon/media_library.ex
@@ -347,4 +347,51 @@ defmodule Beacon.MediaLibrary do
       _ -> :error
     end
   end
+
+  @doc """
+  Returns the contents of an uploaded file as a binary.
+  """
+  @spec read_binary(UploadMetadata.t()) :: {:ok, binary()} | {:error, atom()}
+  def read_binary(metadata) do
+    cond do
+      metadata.node == Node.self() ->
+        File.read(metadata.path)
+
+      Node.connect(metadata.node) == true ->
+        :erpc.call(metadata.node, File, :read, [metadata.path])
+
+      :otherwise ->
+        {:error, :node_not_available}
+    end
+  end
+
+  @doc """
+  Same as `read_binary/1` but raises if the file is not found.
+  """
+  @spec read_binary!(UploadMetadata.t()) :: binary()
+  def read_binary!(metadata) do
+    case read_binary(metadata) do
+      {:ok, binary} -> binary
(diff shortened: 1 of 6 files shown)


lib/beacon/media_library.ex
```diff
@@ -347,4 +347,51 @@ defmodule Beacon.MediaLibrary do
...
+    cond do
+      metadata.node == Node.self() ->
+        File.read(metadata.path)
+
+      Node.connect(metadata.node) == true ->
+        :erpc.call(metadata.node, File, :read, [metadata.path])
+
+      :otherwise ->
+        {:error, :node_not_available}
+    end
```
leandrocp, line 365: Turns out this call is already [optimized by Erlang](https://github.com/erlang/otp/blob/ac7be0d00d9072e28ee9c14d1845f96dc7ed2b25/lib/kernel/src/erpc.erl#L276) so we don't need to check `Node.self()`. Besides that many other errors might raise here (including connection errors) so we'd need to capture and safely return:

```suggestion
    :erpc.call(metadata.node, File, :read, [metadata.path])
  rescue
    error -> {:error, error}
```

lib/beacon/media_library.ex
```diff
@@ -347,4 +347,51 @@ defmodule Beacon.MediaLibrary do
...
+    cond do
+      node == Node.self() ->
+        File.stat(path)
+
+      Node.connect(node) == true ->
+        :erpc.call(node, File, :stat, [path])
+
+      :otherwise ->
+        {:error, :node_not_available}
+    end
```
leandrocp, line 395: Same as the other block code.

leandrocp (changes requested):
Looking good, just a comment to improve the remote call.

leandrocp (approved):

PR #716, opened 2025-01-03
Retry CSS asset loading

leandrocp:


lib/beacon/runtime_css.ex
```diff
@@ -75,9 +77,32 @@ defmodule Beacon.RuntimeCSS do
 
   @doc false
   def current_hash(site) do
+    case get_current_hash(site) do
+      nil ->
+        # retry once if not found
+        Beacon.Loader.load_runtime_css(site)
+        get_current_hash(site)
```
APB9785, line 84: in the case of double-failure, this will return `nil` instead of previously `""`.  this is probably OK but just noting the change

APB9785 (approved):

lib/beacon/runtime_css.ex
leandrocp: I just changed `RuntimeJS` to return `nil` as well and either `nil` or `""` will generate the same path.

PR #717, opened 2025-01-03
Add support for user-defined JS Hooks

### LINKED ISSUE #284: Add support for user-defined client JS hooks
Allow users to pass https://hexdocs.pm/phoenix_live_view/js-interop.html#client-hooks-via-phx-hook to the Beacon livesocket.

The socket is initialized at https://github.com/BeaconCMS/beacon/blob/4dd3b55475eae63fd8594f22510bae6a71c8a9d7/assets/js/beacon.js#L31 and note it does reuse the `LiveView` object from the host app and connects to the same socket path. It's important to reuse the same JS client version to avoid misbehavior.

Ideally the hook code should be stored in the DB and changed at runtime in the admin interface but that imposes some challenges on how to inject the js hook code into the socket.

APB9785:
## Resolves #284


guides/recipes/javascript-interoperability.md
```diff
@@ -0,0 +1,40 @@
+# JavaScript interoperability
+
+Beacon provides support for custom client-side JavaScript via LiveView's [JS Hooks](https://hexdocs.pm/phoenix_live_view/js-interop.html#client-hooks-via-phx-hook).
+
+Hooks are blocks of custom JS code which run when an element in a beacon template reaches a specific
```
leandrocp, line 5:
```suggestion
Hooks are blocks of custom JS code which run when an element in a Beacon template reaches a specific
```

lib/beacon/content/js_hook.ex
```diff
@@ -0,0 +1,50 @@
...
+  use Beacon.Schema
+
+  @type t :: %__MODULE__{}
+
+  schema "beacon_js_hooks" do
+    field :name, :string
+    field :site, Beacon.Types.Site
+
+    field :mounted, :string
+    field :beforeUpdate, :string
```
leandrocp, line 24: Names in Elixir and DB in general are in the snake case format, shouldn't be better to follow that pattern here?
```suggestion
    field :before_update, :string
```

leandrocp (commented):
This is looking good! Just left some minor comments.

lib/beacon/runtime_js.ex
```diff
@@ -58,17 +61,62 @@ defmodule Beacon.RuntimeJS do
...
+  end
+
+  defp format_callback(hook, callback, true) do
+    IO.iodata_to_binary([
+      to_string(callback),
+      "(){",
+      hook |> Map.fetch!(callback) |> String.replace("\n", ""),
+      "},"
+    ])
+  end
```
leandrocp, line 121: You could simplify this code and remove unnecessary calls, something like:

```suggestion
      code =
        [
          format_callback(hook, :mounted, minify?),
          # and ther rest...
        ])

      code =
      if minify? do
        [hook.name, ":{", code, "}"]
      else
        ["    ", hook.name, ": {\n", code, "\n    }"]
      end
      
      IO.iodata_to_binary(code)
    end)
  end

  defp format_callback(hook, callback, minify?)

  defp format_callback(nil = _hook, _callback, _minify?), do: []
 
  defp format_callback(hook, callback, false) do
    [
      "      ",
      to_string(callback),
      "() {\n        ",
      hook |> Map.fetch!(callback) |> String.trim_trailing() |> String.replace("\n", "\n        "),
      "\n      },\n"
    ]
  end

  defp format_callback(hook, callback, true) do
    [
      to_string(callback),
      "(){",
      hook |> Map.fetch!(callback) |> String.replace("\n", ""),
      "},"
    ]
  end
```

leandrocp (commented):
One more

lib/beacon/content/js_hook.ex
APB9785: my intention here was that formatting the callback name would be simple `to_string(name)` since the JS Hook syntax expects camelCase.  I suppose we could switch the field name to snake_case, and then add an additional formatting step like:
```elixir
[head | tail] = String.split(name, "_")
tail = Enum.map(tail, &String.capitalize/1)
Enum.join([head | tail])
```

lib/beacon/content/js_hook.ex
```diff
@@ -0,0 +1,60 @@
...
+
+    js_hook
+    |> cast(attrs, fields)
+    |> validate_required(fields)
+    |> validate_format(:name, ~r/[a-zA-Z_][a-zA-Z0-9_]*/)
+    |> validate_code()
+  end
+
+  @doc false
+  def validate_code(changeset) do
```
leandrocp, line 40: nice

PR #727, opened 2025-01-15
fix: negotiate asset encoding

leandrocp:
Respect accept-enconding request and serve the proper content. Try brotli first then fallback to gzip or deflate (no compression)


APB9785 (approved):
nice 🥳

PR #740, opened 2025-02-04
feat: always generate site endpoints

leandrocp:
- Isolate each site on its own endpoint to avoid conflicts with static paths and other configs
- Use the Proxy Endpoint to generate public URL for sites
- Fix the URL in robots.txt and sitemap.xml (should use the exposed public scheme/port)
- Fix to use Beacon ErrorHTML 
- Update generators to always create the site endpoint and use it in configs
- (Breaking) asset_url -> css_config_url (needs to change admin as well)

### DIFF
--- a/dev.exs
+++ b/dev.exs
@@ -42,9 +42,19 @@ Demo.Repo.start_link()
 Ecto.Migrator.run(Demo.Repo, path, :up, all: true, log_migrations_sql: true)
 Demo.Repo.stop()
 
+Application.put_env(:beacon, DemoWeb.ProxyEndpoint,
+  adapter: Bandit.PhoenixAdapter,
+  check_origin: {DemoWeb.ProxyEndpoint, :check_origin, []},
+  url: [port: 4001, scheme: "http"],
+  http: [ip: {0, 0, 0, 0}, port: 4001],
+  live_view: [signing_salt: "aaaaaaaa"],
+  secret_key_base: String.duplicate("a", 64),
+  server: true
+)
+
 Application.put_env(:beacon, DemoWeb.Endpoint,
   adapter: Bandit.PhoenixAdapter,
-  http: [ip: {0, 0, 0, 0}, port: 4001],
+  http: [ip: {0, 0, 0, 0}, port: 4100],
   server: true,
   live_view: [signing_salt: "aaaaaaaa"],
   secret_key_base: String.duplicate("a", 64),
@@ -87,12 +97,21 @@ defmodule DemoWeb.Router do
   end
 end
 
+defmodule DemoWeb.ProxyEndpoint do
+  use Beacon.ProxyEndpoint,
+    otp_app: :beacon,
+    session_options: [store: :cookie, key: "_beacon_dev_key", signing_salt: "pMQYsz0UKEnwxJnQrVwovkBAKvU3MiuL"],
+    fallback: DemoWeb.Endpoint
+end
+
 defmodule DemoWeb.Endpoint do
   use Phoenix.Endpoint, otp_app: :beacon
 
   @session_options [store: :cookie, key: "_beacon_dev_key", signing_salt: "pMQYsz0UKEnwxJnQrVwovkBAKvU3MiuL"]
 
-  socket "/live", Phoenix.LiveView.Socket, websocket: [connect_info: [session: @session_options]]
+  # TODO: add this function in the `gen.site` generator
+  def proxy_endpoint, do: DemoWeb.ProxyEndpoint
+
   socket "/phoenix/live_reload/socket", Phoenix.LiveReloader.Socket
 
   plug Phoenix.LiveReloader
@@ -1246,7 +1265,8 @@ Task.start(fn ->
     Demo.Repo,
     {Phoenix.PubSub, [name: Demo.PubSub]},
     {Beacon, sites: [dev_site, dy_site]},
-    DemoWeb.Endpoint
+    DemoWeb.Endpoint,
+    DemoWeb.ProxyEndpoint
   ]
 
   {:ok, _} = Supervisor.start_link(children, strategy: :one_for_one)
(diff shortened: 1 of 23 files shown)


lib/beacon/web/sitemap/sitemap_index.xml.eex
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
-  <%= for url <- @urls do %><sitemap>
+<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"><%= for url <- @urls do %>
+  <sitemap>
```
APB9785, line 3: good catch

dev.exs
```diff
@@ -1246,7 +1265,8 @@ Task.start(fn ->
     Demo.Repo,
     {Phoenix.PubSub, [name: Demo.PubSub]},
     {Beacon, sites: [dev_site, dy_site]},
```
APB9785, line 1267: should these sites get their own endpoint now instead of using `DemoWeb.Endpoint`?

dev.exs
```diff
@@ -87,12 +97,21 @@ defmodule DemoWeb.Router do
...
+    session_options: [store: :cookie, key: "_beacon_dev_key", signing_salt: "pMQYsz0UKEnwxJnQrVwovkBAKvU3MiuL"],
+    fallback: DemoWeb.Endpoint
+end
+
 defmodule DemoWeb.Endpoint do
   use Phoenix.Endpoint, otp_app: :beacon
 
   @session_options [store: :cookie, key: "_beacon_dev_key", signing_salt: "pMQYsz0UKEnwxJnQrVwovkBAKvU3MiuL"]
 
-  socket "/live", Phoenix.LiveView.Socket, websocket: [connect_info: [session: @session_options]]
```
APB9785, line 95: with this removed from the existing Endpoint, will this break pages outside of beacon?  can the ProxyEndpoint work for non-beacon pages?

dev.exs
```diff
@@ -87,12 +97,21 @@ defmodule DemoWeb.Router do
...
+  # TODO: add this function in the `gen.site` generator
+  def proxy_endpoint, do: DemoWeb.ProxyEndpoint
```
APB9785, line 113: If this is supposed to represent the "default" Endpoint that Phoenix gives you, then this should go in `mix beacon.gen.proxy_endpoint` in the `update_existing_endpoints/3` because that's the only place we touch that module.

lib/beacon/router.ex
```diff
@@ -185,6 +185,9 @@ defmodule Beacon.Router do
   Then Beacon will reference both of those sitemaps in the top-level index:
     * `my_domain.com/sitemap_index.xml`
 
+  Note that `beacon_sitemap_index` will include the sitemap URL of all mounted sites
+  in the router, so that macro should be at the root and not duplicated.
+
```
APB9785, line 190: what does it mean that it should be at the root?  is it still OK for e.g.
```elixir
scope "/admin", as: :beacon_live_admin do
  pipe_through [:admin_browser, ...]

  get "/session/sign_out", SessionController, :sign_out

  beacon_live_admin "/"
end

scope "/" do
  pipe_through [:browser, ...]
  beacon_site "/", site: :dockyard_com
end
```

dev.exs
leandrocp: Ideally yes to simulate the same config used in a real app. I created the Proxy endpoint but forgot to create the site endpoints.

dev.exs
leandrocp: This endpoint is supposed to be one of the site endpoints, probably should rename it to `DevEndooint` and create a new `DyEndpoint`. No need to have an "app endooint" in dev.exs tho because we  wouldn't use it for anything.

dev.exs
leandrocp: It's the other way around actually. From the site endpoint we need to know the proxy endpoint in order to fetch the scheme and port to generate URLs. I did include that function mostly to avoid adding a new required option in `Beacon.Config`. I believe that's okay because that endpoint is generated by `gen.site` and used only by beacon sites. Sounds good?

APB9785: I think the confusion was, I thought this was supposed to represent the "default" phoenix endpoint, and not a generated beacon endpoint.  So my comment is no longer relevant 🙂

lib/beacon/router.ex
APB9785: ok just checking there wasn't an additional meaning that needed documentation

APB9785 (approved):

lib/beacon/router.ex
leandrocp: Wait... giving a second thought we can't render all sites in the sitemap index. Giving 2 different scopes on different `:host` it should include only the sites within that scope.

APB9785: that would mean multiple sitemap indexes, right?  we don't currently have a way to do that 🤔

leandrocp: Eg: 

```
scope host: foo.com
  beacon_site :site_a  
  beacon_site :site_b 
  sitemap_index   

scope host: bar.com
  beacon_site :site_c 
  sitemap_index
```

Should foo.com/sitemap_index.xml include site_c?

APB9785: I think in order to properly implement this, we need some data model "in-between" 
```
[otp_app] --- [ ??? ] --- [Beacon :site]
```
where that `???` would represent a "host"

leandrocp: `sitemap.xml` and `robots.txt` scoped by site are fine, served at `{host}/{prefix}/{file}`, but the sitemap index it seems like should only include the sites running in the same scope where it's defined or have an option to define which sites should be included in the index.

So either an option like:

```
scope host: foo.com
  beacon_site :site_a  
  beacon_site :site_b 
  sitemap_index [:site_a, :site_b]

scope host: bar.com
  beacon_site :site_c 
  sitemap_index [:site_c]
```

_Where the default could be "all" sites._

Or introspect the router to accumulate sites per scope, similar to how we accumulate sites in `__beacon_sites__` in the router.

Although that's not responsibility of this PR.

PR #761, opened 2025-02-13
Add `--endpoint` option for `beacon.gen.site`

### LINKED ISSUE #759: `beacon.gen.site` option for customizing Endpoint module
We have an endpoint which varies from the default naming scheme.  In order to use the `beacon.gen.site` task with idempotence (i.e. to upgrade a site or reset it to default configuration), an option should be exposed for us to declare the endpoint name/prefix explicitly.

APB9785:
## Resolves #759


lib/mix/tasks/beacon.gen.site.ex
```diff
@@ -506,8 +507,8 @@ if Code.ensure_loaded?(Igniter) do
     end
 
     @doc false
-    def new_endpoint_module!(igniter, site) when is_atom(site) do
-      suffix = new_endpoint_module!(site)
+    def site_endpoint_module!(igniter, site) when is_atom(site) do
+      suffix = site_endpoint_module!(site)
```
leandrocp, line 511:
```suggestion
      suffix =       site
      |> to_string()
      |> String.split(~r/[^[:alnum:]]+/)
      |> Enum.map_join("", &String.capitalize/1)
      |> Kernel.<>("Endpoint")
```

And then you can delete `site_endpoint_module!/1`

lib/mix/tasks/beacon.gen.site.ex
```diff
@@ -101,6 +104,7 @@ if Code.ensure_loaded?(Igniter) do
 
       port = Keyword.get_lazy(options, :port, fn -> Enum.random(4101..4999) end)
       secure_port = Keyword.get_lazy(options, :secure_port, fn -> Enum.random(8444..8999) end)
+      site_endpoint = Keyword.get_lazy(options, :endpoint, fn -> site_endpoint_module!(igniter, site) end)
```
leandrocp, line 107: Values given from args are in binary so you need to convert to a proper module:

```suggestion
      site_endpoint = Keyword.get_lazy(options, :endpoint, fn -> site_endpoint_module!(igniter, site) end)
      site_endpoint = Module.concat([site_endpoint])
```

lib/mix/tasks/beacon.gen.site.ex
```diff
@@ -33,6 +33,8 @@ defmodule Mix.Tasks.Beacon.Gen.Site.Docs do
     * `--host` (optional) - If provided, site will be served on that host.
     * `--port` (optional) - The port to use for http requests. Only needed when `--host` is provided.  If no port is given, one will be chosen at random.
     * `--secure-port` (optional) - The port to use for https requests. Only needed when `--host` is provided.  If no port is given, one will be chosen at random.
+    * `--endpoint` (optional) - The name of the Endpoint Module for your site. If not provided, a default will be generated based on the `site`.
```
leandrocp, line 36: Depending on the name, Igniter will try to create files in a path that users may not expect. For eg:

```
--endpoint DockYardWeb.Endpoint
```

Creates the file `lib/dock_yard_web/endpoint.ex`

Which is expected because it follows the Phoenix convention. Are we just follow that pattern or find a way to overwrite the path name as well?

leandrocp (changes requested):

leandrocp (approved):

Issue #764, opened 2025-02-13
beacon.gen.site fails when site endpoint already exists


APB9785:
### Actual behavior

Running `mix beacon.gen.site` fails if the site's endpoint already exists (such as from running the task previously)

```
Issues:

* lib/my_app_web/site_endpoint.ex: File already exists
```

### Expected behavior

Generator runs and resets that endpoint module to its default state

PR #765, opened 2025-02-13
fix(beacon.gen.site): don't fail on pre-existing endpoint module or remove extra beacon runtime config

### LINKED ISSUE #763: beacon.gen.site removes Beacon site config
### Actual behavior

Running `mix beacon.gen.site` to upgrade/reset an existing site removes additional `Beacon.Config` that might be defined in `runtime.exs`

### Expected behavior

Running that generator only resets necessary keys and doesn't touch extra config

APB9785:
## Resolves #763 and #764 

* Now overwrites site endpoint if it already exists, instead of failing the task (#764)
* Don't remove extra Beacon.Config the user may have set for the site in `runtime.exs` (#763)

### DIFF
--- a/lib/mix/tasks/beacon.gen.site.ex
+++ b/lib/mix/tasks/beacon.gen.site.ex
@@ -386,7 +386,6 @@ if Code.ensure_loaded?(Igniter) do
         igniter,
         new_endpoint_module!(igniter, site),
         contents,
-        # maybe need to wrap contents?
         fn zipper -> {:ok, Igniter.Code.Common.replace_code(zipper, contents)} end
       )
     end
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/beacon.gen.site.ex
```diff
@@ -344,51 +339,55 @@ if Code.ensure_loaded?(Igniter) do
...
+      plug Plug.Head
+      plug Plug.Session, @session_options
+      plug #{inspect(web_module)}.Router
+      """
+
+      Igniter.Project.Module.find_and_update_or_create_module(
+        igniter,
+        new_endpoint_module!(igniter, site),
+        contents,
+        # maybe need to wrap contents?
```
leandrocp, line 389: What do you mean by wrapping contents?

leandrocp (approved):

PR #779, opened 2025-02-25
feat: introduce `url_host/2` and `:hosts` Beacon.Config

leandrocp:
Used to define the `:hosts` endpoint config dynamically, so the same site can be served from multiple domains even if multiple sites are served in the same prefix.

More docs to be added along with adjustments to generators in a following PR.

### DIFF
--- a/lib/beacon/config.ex
+++ b/lib/beacon/config.ex
@@ -63,6 +63,20 @@ defmodule Beacon.Config do
   @type mode :: :live | :testing | :manual
 
   @typedoc """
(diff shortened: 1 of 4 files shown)


APB9785 (approved):
Nice 🚀

leandrocp:
Closed in favor of https://github.com/BeaconCMS/beacon/pull/786 which is a bit less flexible (doesn't support multiple hosts for the same env) but is a simple solution that doesn't require adding a new config.

PR #782, opened 2025-02-26
adds customizable asset delete callbacks

ddink:
### **What will it allow you to do that you can't do today?**

It will allow users to customize a `soft_delete/1` callback that's used when `Beacon.MediaLibrary.soft_delete/1` gets called.

### **Why do you need this feature and how will it benefit other users?**

This feature is necessary to be able to programmatically delete stored assets where they are stored when utilizing custom provider modules for the media library's asset upload functionality.

[Previous PR ](https://github.com/BeaconCMS/beacon/pull/601)


leandrocp:
Thanks!

PR #783, opened 2025-02-26
fix(gen.site): multiple routers prompt

leandrocp:


tomciopp:
This should probably use `Keyword.get/3` instead of an `||` condition.

leandrocp:
`Keyword.get/3` has a small flaw that still permits `nil` values:

```
iex> Keyword.get([label: nil], :label, "default")
nil

iex> Keyword.get([label: nil], :label) || "default"
"default"
```

We want the later :)

APB9785 (approved):

leandrocp:
@tomciopp do you mind giving it a try?

```
{:beacon, github: "BeaconCMS/beacon", override: true}
```

If that works as expected and no other bugs are found we can release a patch version

tomciopp:
@leandrocp Everything looks good with those changes!

leandrocp:
@tomciopp v0.4.2 just released, thanks!

Issue #776, opened 2025-02-24
Beacon.Registry GenServer terminates when running Phoenix project with custom Ecto Repo timestamp columns in database


sltong:
### Environment

* Elixir version (elixir -v): Elixir 1.18.2 (compiled with Erlang/OTP 27) 
* Phoenix version (mix deps): 1.7.20
* Phoenix LiveView version (mix deps): 1.0.4
* Beacon version (mix deps): 0.4
* Operating system: macOS Sequoia (15.3.1)
* Browsers on which you attempted to reproduce this bug (the more the merrier): N/A

### Minimal reproduction steps
1. Install the Phoenix and Igniter installers: `mix archive.install hex phx_new && mix archive.install hex igniter_new`
2. Generate a Phoenix project: `mix phx.new custom_timestamps_opts`
3. Navigate to the project's root directory: `cd custom_timestamps_opts`
4. Add a configuration for the project's Ecto Repo with a custom `inserted_at` timestamp field (for migrations):
```
config :custom_timestamps_opts, CustomTimestampsOpts.Repo,
  migration_timestamps: [inserted_at: :created_at]
```
5. Install Beacon and use the default answers to prompts: `mix igniter.install beacon,beacon_live_admin --beacon.site custom_timestamps_opts`
6. Create the project database: `mix ecto.create`
7. Run the Phoenix server: `mix phx.server`

### Actual behavior

The `Beacon.Registry` GenServer will terminate with the following output:

```
[info] Beacon.Boot booting site custom_timestamps_opts
[debug] QUERY ERROR source="beacon_page_snapshots" db=0.4ms queue=2.0ms idle=0.7ms
[...]

PR #793, opened 2025-03-11
Hardcode "inserted_at" & "updated_at" in Ecto migrations & schemas.

sltong:
Resolves issue #776

Hardcode the "inserted_at" and "updated_at" names for timestamps options in Ecto migrations and schemas. As Beacon is configured to use a host application's Ecto Repo, querying for those columns by their default names may fail if the host application defines custom column names (e.g. "created_at" in place of "inserted_at").


leandrocp:
Thanks! 🚀

PR #801, opened 2025-03-24
Update tailwind package to 0.3

cibernox:
This is necessary so the live_admin, which depends on 0.3.1, can resolve dependencies.


leandrocp:
There are 2 issues with bumping `:tailwind` version:

1. That's a breaking change that impose the same change to users' projects. Better to keep `~> 0.2` for now (which doesn't block users to use 0.3 in their projects).
2. Need more testing since v0.3 assumes tailwind v4 which we don't support yet so we need to make sure it works well.

Issue #821, opened 2025-04-24
Match error on rendering error pages ErrorHTML


javierg:
### Environment

* Elixir version (elixir -v): 1.17.2
* Phoenix version (mix deps): 1.7.21
* Phoenix LiveView version (mix deps): 1.0.10
* Beacon version (mix deps): 0.5.1
* Operating system: OSX 15.4
* Browsers on which you attempted to reproduce this bug (the more the merrier): Chrome

### Actual behavior

Site will not render configured error page on beacon_live_admin and raise a Match error such as:

```
[warning] failed to render error page for 404.html, fallbacking to default Phoenix error page

Got:

%MatchError{term: {Beacon.Web.PageLive, [action: :path, router: MyAppWeb.Router], %{extra: %{session: {Beacon.Router, :session, [:home, %{}]}, root_layout: {Beacon.Web.Layouts, :runtime}, on_mount: []}, name: :beacon_home}}}
```

### Expected behavior

Error pages should be rendered without any raised errors

PR #822, opened 2025-04-26
Fix:#821 Supports private phoenix_live_view session as tuple and map

javierg:
This is a proposal solution for https://github.com/BeaconCMS/beacon/issues/821

What Is aw in my dev environment is that `conn.private.phoenix_live_view` extra session value is set as 3 items tuple, instead of a map, as expected. Not sure if this changed some where in the past and is actually expected.
I can't figure it out clearly what was going on, and why the session is expected as a map, every test I did, returned as a tuple, not sure if this was a change on LiveView itself, I didn't research much from that side.

Hopefully you have time to review this and clarify if needed (I am using this branch for my working site as I need this custom error pages to work), If there is an alternative, I will appreciate it

### DIFF
--- a/lib/beacon/web/controllers/error_html.ex
+++ b/lib/beacon/web/controllers/error_html.ex
@@ -8,7 +8,7 @@ defmodule Beacon.Web.ErrorHTML do
 
   @doc false
   def render(<<status_code::binary-size(3), _rest::binary>> = template, %{conn: conn}) do
-    {_, _, %{extra: %{session: %{"beacon_site" => site}}}} = conn.private.phoenix_live_view
+    site = conn.private.phoenix_live_view |> fetch_session_site()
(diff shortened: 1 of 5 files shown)


lib/beacon/web/controllers/error_html.ex
```diff
@@ -29,4 +29,11 @@ defmodule Beacon.Web.ErrorHTML do
     Logger.warning("could not find an error page for #{template}, fallbacking to default Phoenix error page")
     Phoenix.Controller.status_message_from_template(template)
   end
+
+  defp fetch_session_site({_, _, %{extra: %{session: site_session}}}) do
```
javierg, line 33: is there a place to put shared code like this?

javierg:
I cant fix the Github action weird issue that can't pull from my forked repo. Will try closing the PR and opening a new one.

PR #851, opened 2025-12-12
Adjust according to breaking changes in MDEx 0.9

velzevur:
MDEx introduced breaking changes in [v0.9](https://github.com/leandrocp/mdex/blob/main/CHANGELOG.md#090-2025-09-22). Without it one can not use latest versions of the dep. This PR makes it compatible with it.


alexe-argo (approved):
This looks good and accounts for the security vulnerabiltiies in earlier versions

